PR #20, opened 2019-02-01
Create defflow macro

caioaao:
The most important part of this macro is to add metadata to the var so we know it points to a flow. This is useful for writing a runner later (I'm planning on implementing a [lambdaisland/kaocha](/lambdaisland/kaocha) extension for it)

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -142,3 +142,16 @@
   "Returns all flows defined with `defflow`"
   [ns]
   (->> ns ns-interns vals (filter (comp ::test meta))))
+
+(defn run-test*
+  [v]
+  (let [{::keys [cleanup initialize]} (meta v)
+        initial-state (initialize)
+        [_ final-state :as result] (run @v initial-state)]
+    (cleanup final-state)
+    result))
+
+(defmacro run-test
+  "Runs test `test-name`defined with `deftest`"
+  [test-name]
+  `(run-test* (var ~test-name)))
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -4,7 +4,7 @@
             [cats.core :as m]
             [cats.data :as d]
             [cats.monad.exception :as e]
-            [clojure.test :refer :all]
+            [clojure.test :as ctest :refer [is]]
             [matcher-combinators.test]
             [midje.checking.core :refer [extended-=]]
             [midje.sweet :refer :all]
@@ -50,11 +50,6 @@
     pop-meta
     (m/return ret#)))
 
-(defmacro defflow
-  [sym descr & flows]
-  `(def ~(vary-meta sym assoc ::test true)
-     (flow ~descr ~@flows)))
-
 (defn retry
   "Tries at most n times, returns a vector with true and first element that succeeded
   or false and result of the first try"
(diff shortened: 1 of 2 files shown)


ericvm:
I think that a test should actually be a specific run of a flow with some specific initial-state.

I'm not sure how this would be implemented, though :thinking:

caioaao:
hm, makes sense. probably better to create `deftest` that receives a function to initialize state and another for  cleanup (like test fixtures)

src/state_flow/core.clj
```diff
@@ -124,3 +126,32 @@
...
+(defn ns->tests
+  "Returns all flows defined with `defflow`"
+  [ns]
+  (->> ns ns-interns vals (filter (comp ::test meta))))
+
+(defn run-test*
+  [v]
+  (let [{::keys [cleanup initialize]} (meta v)
+        initial-state (initialize)
+        [_ final-state :as result] (run @v initial-state)]
```
ericvm: One idea that I have is to make tests that require `matcher-combinators` library only and generate a report that is stored on the `final-state`. Then you could drop dependency from clojure.test and midje and register the flow run itself as a test run. Does that make sense?

src/state_flow/core.clj
```diff
@@ -124,3 +126,32 @@
   "Transform a flow step into a state transition function"
   [flow]
   (fn [s] (state/exec flow s)))
+
+(defmacro deftest
+  "Define a test to be run.
+
+  Receives a function `initialize-state`, a function with no
+  arguments that returns the initial state."
+  [sym descr initialize-state cleanup & flows]
```
ericvm: I think it would be more future proof to require a parameter map with optionals, it is better to extend further this way.

For instance:
- descr might be optional because you are assigning a symbol and the default description can be `(str sym)`.
- Default cleanup might be `identity`
- Default initialize should be `{}`, also the state is expected to be a map, might be nice to assert this

You could (but should you?) do initializing and cleanup inside the flow itself. So it might be good to make it optional.

ericvm: Another thing that it is useful to add: the default flow runner. Currently you can run with `run` or `run!`. Having `run!` as default might be good good to avoid silently ignoring exceptions.

src/state_flow/core.clj
caioaao: I thought of doing it that way, but I don't like mixing state with the report. one could want to override the whole state in the middle of a flow, and then we'd lose the report

src/state_flow/core.clj
caioaao: yeah, I saw `run!` but didn't think it was a good idea to use `clojure.logging`. I like the idea of having `run` returning the reports and delegating presentation to the runner

src/state_flow/core.clj
ericvm: The way it works right now, state-flow already uses the state internally. Flow description is stored on `[:meta :description]`. You do not mix because all the internals should be inside `:meta` keyword. Normal operation shouldn't involve wiping out the state, otherwise things will not work as expected.

Avoiding using the state for descriptions or test-reports would require a lot of macro voodoo magic. How would you do that?

caioaao: maybe use a spec for the state like `(s/keys :req [::meta])` then and always check if state conforms to it after a run. this way we can make these types of bugs more obvious

ericvm:
done on https://github.com/nubank/state-flow/pull/23

Issue #21, opened 2019-02-02
Short circuit on failed assertion


caioaao:
I would expect a failed assertion (i.e.: failed `verify` or `match?`) would short circuit a flow. This is not the case and as of now we don't have an easy way of asserting that it happened when running a flow.

philomates:
One thing to keep in mind:
I believe the way that `kaocha` uses `clojure.test` counters, as well as [`selvage`](https://github.com/nubank/selvage/blob/aabd8c2a3fbc57fe0dcee262e447502578d3308b/src/selvage/test/flow.clj#L83), which copied `kaocha`, is that it has issues with test runners that run tests in parallel. I think this is why `bat-test` doesn't play well with `selvage`'s `defflow`.
I want to try to fix `selvage` to run tests in parallel at some point so I'll try to do it in a way that can be used here as well

ericvm:
I don't see a strong reason to short circuit. I actually see reasons for not short circuiting. Supposing you do an action that has several consequences, than you make several assertions for each of the consequences. These assertions are independent and it is meaningful to know which of them passed or failed. If the flow didn't fail catastrophically (with an exception) it probably should continue. The spice must flow.

But I'm all open for having this behavior configurable. And also decoupling state-flow from test assertions

caioaao:
I agree with the multiple assertion, but not steps. If an assertion fails between two steps, the step didn't work as expected and imo it doesn't make sense to keep running the next steps (assuming a flow is exactly for actions that depend on the effect of previous actions)

dchelimsky:
>  If an assertion fails between two steps, the step didn't work as expected and imo it doesn't make sense to keep running the next steps

To the runner, flows, steps, assertions, etc, are all the same thing, so there is no way to differentiate between a step and an assertion. Also, since you can arbitrarily nest flows inside assertions inside flows, there is really no way to identify when "an assertion fails between two steps"

I think the simplest thing is to either short circuit on error or not (configurable - default TBD), and runtime exceptions should be treated the same as failed assertions.

philomates:
Hey, so the `run*` function was extended to take a `fail-fast?` option to achieve this. This has been released in `5.6.0` via https://github.com/nubank/state-flow/pull/138

PR #22, opened 2019-02-04
Add `assert` macro

caioaao:
Fixes nubank/state-flow#21

The `assert` macro receives a body and, if this body doesn't return a truthy value, it will throw an exception with the details to short circuit the flow.

TBH this is a hacky workaround. I'm experimenting with using an either monad instead of the exception monad and detect test errors differently, but I still don't have a working solution. I think it might be an opportunity to do what we discussed before and also return the test report when running it.

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -1,5 +1,6 @@
 (ns state-flow.core
-  (:refer-clojure :exclude [run!])
+  (:refer-clojure :exclude [run!]
+                  :rename {assert core-assert})
   (:require [cats.context :as ctx]
             [cats.core :as m]
             [cats.data :as d]
@@ -104,8 +105,8 @@
 
 (defn run
   [flow initial-state]
-  (assert (state/state? flow) "First argument must be a State Monad")
-  (assert (map? initial-state) "Initial state must be a map")
+  (core-assert (state/state? flow) "First argument must be a State Monad")
+  (core-assert (map? initial-state) "Initial state must be a map")
   (state/run flow initial-state))
 
 (defn run!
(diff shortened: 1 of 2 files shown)


src/state_flow/core.clj
```diff
@@ -124,3 +125,16 @@
...
+  exception."
+  {:style/indent 1}
+  [desc & body]
+  `(flow ~desc
+     [full-desc# (get-description)]
+     (let [ret# ~@body]
+       (or (and ret# (m/return ret#))
+           (throw (ex-info (str "Flow failed at assertion " full-desc#)
+                           {:reason ::assertion-failed
+                            :form '~body}))))))
```
ericvm: You should take care to diffrentiate what happens on definition time and running time. It seems like this assertion is being done when a flow is defined. This would not work if you are asserting against an internal state at runtime.

The way to go for running arbitrary functions at flow runtime is to use `state/wrap-fn` (or at least that is what I came up with).

But probably the best thing to do here is to accept a flow as an argument instead of arbitrary body.

```clojure
(defn assert
  "Defines a step that throws an
  exception if return value is not a `truthy`"
  {:style/indent 1}
  [desc flow-step]
  (flow desc
     [full-desc (get-description)
      ret          flow-step]
     (or (and ret (m/return ret))
          (throw (ex-info (str "Flow failed at assertion " full-desc#)
                           {:reason ::assertion-failed
                            :form '~}))))))
```

The reason `verify` and `match?` are macros is to get the correct line number when things fail. Probably it would be nice for this to be a macro too, but maybe it's not that important. :thinking:

caioaao:
I'd forgotten about this PR until I opened the other one :man_facepalming: #45 should address this issue

PR #23, opened 2019-03-25
Improve support for clojure.test

ericvm:
Based on https://github.com/nubank/state-flow/pull/20

* `match?` macro now expands to a clojure.test `testing` and uses `meta` variables from the position it has been written, therefore giving correct line number when a test fails.
* Optional parameters for setting initial-state, cleanup and flow runner

Usage will look something like this:

```clojure
(defflow my-flow
  (match? "simple test" 1 1)
  (match? "embeds" {:a 1 :b 2} {:a 1}))
```
Or with custom parameters:

```clojure
(defflow my-flow {:init aux.init! :runner (comp run! s/with-fn-validation)}
  (match? "simple test" 1 1)
  (match? "simple test 2" 2 2))
```

```clojure
(defflow my-flow {:init (constantly {:value 1
                                     :map {:a 1 :b 2}})}
  [value (state/gets :value)]
  (match? "value is correct" value 1)
  (match? "embeds" (state/gets :map) {:b 2}))
```

example flow migration: https://github.com/nubank/arnaldo/pull/169/files

### DIFF
--- a/test/state_flow/cljtest_test.clj
+++ b/test/state_flow/cljtest_test.clj
@@ -35,6 +35,13 @@
                  {:value {:a 2 :b 5}
                   :meta {:description []}})))
 
+  (fact "works for failure cases"
+    (let [val {:value {:a 2 :b 5}}]
+      (state-flow/run (cljtest/match? "contains with monadic left value" (state/gets :value) (matchers/equals {:a 1 :b 5})) val)
+      => (d/pair {:a 2 :b 5}
+                 {:value {:a 2 :b 5}
+                  :meta {:description []}})))
+
   (fact "works with matcher combinators in any order"
     (let [val {:value [1 2 3]}]
       (state-flow/run (cljtest/match? "contains with monadic left value" (state/gets :value) (matchers/in-any-order [1 3 2])) val)
--- a/test/state_flow/cljtest_test.clj
+++ b/test/state_flow/cljtest_test.clj
@@ -4,9 +4,8 @@
             [cats.monad.state :as state]
             [clojure.test :as ctest]
             [matcher-combinators.matchers :as matchers]
-            [matcher-combinators.midje :refer [match]]
             [midje.sweet :refer :all]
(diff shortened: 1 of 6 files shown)


test/state_flow/cljtest_test.clj
```diff
@@ -0,0 +1,111 @@
...
+      => (d/pair {:a 2 :b 5}
+                 {:value {:a 2 :b 5}
+                  :meta {:description []}})))
+
+  (fact "works with matcher combinators in any order"
+    (let [val {:value [1 2 3]}]
+      (state-flow/run (cljtest/match? "contains with monadic left value" (state/gets :value) (matchers/in-any-order [1 3 2])) val)
+      => (d/pair [1 2 3]
+                 {:value [1 2 3]
+                  :meta {:description []}}))))
```
lucascs: can you check for failures as well?

failure messages are more important than successes for test frameworks.

ericvm: I added failure cases. Problem with that is that it prints some failure info because of this, but tests pass.

test/state_flow/cljtest_test.clj
```diff
@@ -50,70 +49,18 @@
...
+          (state-flow.core/run*
+           {}
+           (state-flow.core/flow (clojure.core/str my-flow) (cljtest/match? 1 1)))))
+  (fact "defines flow with optional parameters"
+    (macroexpand-1 '(defflow my-flow {:init (constantly {:value 1})} (cljtest/match? 1 1)))
+      => '(clojure.test/deftest
+            my-flow
+            (state-flow.core/run*
+             {:init (constantly {:value 1})}
+             (state-flow.core/flow (clojure.core/str my-flow) (cljtest/match? 1 1))))))
```
lucascs: I'd use something like 
```clj
(state-flow.core/flow "my-flow" (cljtest/match? 1 1))
```
It might have an easy way to unquote the `str` call.

PR #29, opened 2019-05-27
Timeout params

ericvm:
Add optional parameters to `match?`, making it possible to tweak times-to-try and sleep-time of test probing


danilobuiatti:
pretty nice!

PR #32, opened 2019-08-08
wrapper-fn defflow parameter to allow for with-redefing a whole flow

philomates:
I want to redef `call-async` for an entire defflow, just like can be done for `midje` flows:
https://github.com/nubank/purgatory/blob/45296d3e6d5111086f4e0606bdca879b32a56af7/postman/postman/delinquent.clj#L117

It isn't currently possible, but the same issue was solved in `selvage` by introducing a `:wrapper-fn` parameter to `defflow` and wrapping the whole `run` in that wrapper.

If people like this approach I'll
 - write docs in the readme
 - update the `defflow` docstring to mention `:wrapper-fn`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [1.15.0]
+- add `:wrapper-fn` parameter to `defflow` parameters, which allows you to wrap a whole flow with things like `with-redefs`.
+
 ## [1.14.0]
 - Allow for empty flows
 
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject nubank/state-flow "1.14.0"
+(defproject nubank/state-flow "1.15.0"
 
   :description "Postman-like integration testing with composable flows"
   :url "https://github.com/nubank/state-flow"
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -44,6 +44,9 @@
   [name & forms]
   (let [[parameters & flows] (if (map? (first forms))
                                forms
-                               (cons {} forms))]
+                               (cons {} forms))
+        wrapper              (or (:wrapper-fn parameters)
+                                 (fn [flow] (flow)))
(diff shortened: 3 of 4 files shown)


ericvm:
What are the kinds of problems you want to solve, can you give an example? Because I always aim for avoiding `with-redefs` at all costs.

I don't have a problem with offering the possibility, but I'm afraid people might abuse it.

philomates:
I often use them to redefine config functions for an entire flow. Not sure if there is a less invasive way to do that. And in our datomic log extractor service, which has flows that extract from a local datomic, we have to redef/rebind a lot of code to adjust retry and batching logic. In the general case I don't think it isn't needed, but for complex flows it can be very helpful. Also, if you want to implement finalizers, like https://github.com/nubank/common-test/pull/191, it can be done using `wrapper-fn` + a try-catch

ericvm:
If you want to update a system dependency, I recommend using this: https://github.com/nubank/state-flow-helpers/blob/master/src/state_flow/helpers/core.clj#L69

Also, there is the problem of thread-safety with `with-redefs`.

I was actually trying to do a wrapper for a general flow instead of a `defflow`, but didn't have much success with this, but if you do it for a flow instead of a defflow I think it would be better.

My goal was to make sure that all async threads were finished before proceeding to the next step (without having to make the processes synchronous). This would be useful for the `with-delta` function that fetches the delta of interaction that happened in a step, it would be possible to fetch a correct delta even if there are asynchronous processes.

I'm ok with merging the PR, but I'm wondering if there is a better way.

ericvm:
```clojure
(defn wrap-flow wrapper-fn flow
  (m/do-let
    [world (state/get)
     [ret world'] (wrapper-fn (run! flow world))]]
     (state/put world')
     (m/return ret)))
```

I think the general code would be something like this. I think it could work.

ericvm:
@philomates I came up with this that seems to be working: https://github.com/nubank/state-flow/pull/33
See if it makes sense

philomates:
closing in favor of the approach being explored in https://github.com/nubank/state-flow/pull/33

PR #33, opened 2019-08-13
wrap-with macro

sovelten:


src/state_flow/core.clj
```diff
@@ -131,6 +131,14 @@
     (cleanup final-state)
     result))
 
+(defmacro wrap-with
+  [wrapper-fn flow]
+  `(m/do-let
+    [world# (state/get)
+     pair#  (m/return (~wrapper-fn (run! ~flow world#)))]
+    (state/put (second pair#))
+    (m/return (first pair#))))
```
philomates: how about 
```clojure
(defmacro wrap-with
  [wrapper-fn & flows]
  `(m/do-let
    [world# (state/get)
     pair#  (m/return (~wrapper-fn (run! (m/do-let ~@flows) world#)))]
    (state/put (second pair#))
    (m/return (first pair#))))
```
to allow for wrapping multiple flows (https://github.com/nubank/ouroboros/pull/48/commits/121b1c1fa876f38cb7c7d8297de3af37e4b21769#diff-cc666e97deaf80145512a5b95db23dd6R115)

philomates:
This seems like a good way to tackle the issues I was trying to address in  https://github.com/nubank/state-flow/pull/32 without adapting the core logic. It does require users to write macros whenever they want to this type of expressiveness. Maybe this is a good thing because it will deter folks and we strive to cultivate a set of wrappers that tackle most needs.
Of course tackling issues via macros always feels a little dirty so I wonder if there is some unforeseen thing that will bite us. Given that the change is small and contained I guess the blast radius of this is quite small

src/state_flow/core.clj
ericvm: You can always do `(wrap-with wrapper (flow "" &flows))`

I'm still thinking about the runner (we should be able to supply one instead of default) and also the extensibility of the macro. If you do variadic arguments it is difficult to extend it later.

philomates: It might be a total hack but you could store the runner as metadata of the flow and have the `wrap-with` look up the runner of the flow arg and use it:

```clojure
(defmacro wrap-with [flow]
  `(let [runner# (-> ~flow meta :runner)]
     (runner# ~flow)))

(def f ^{:runner (fn [flow] (flow 1))}
  (fn [x] (inc x)))
```

dchelimsky: What about letting the `flow-or-flows` argument be a single flow or a seq of flows e.g.

``` clojure
;; just for discussion - I didn't run this
(defmacro wrap-with
  [wrapper-fn flow-or-flows]
  `(m/do-let
    [world# (state/get)
     pair#  (m/return 
             (~wrapper-fn
              (run!
               (m/do-let ~@(if (sequential? flow-or-flows) flow-or-flows [flow-or-flows]) world#)))]
    (state/put (second pair#))
    (m/return (first pair#))))
```

This way you can support multiple flows now and extend it later.

Alternatively, can't you support multiple flows by nesting them in a single flow? e.g.

``` clojure
(wrap-with f (flow one-flow another-flow))
```

sovelten: @dchelimsky Thanks for the comment!
1. Yes, passing a sequence for multiple flows will allow for extensibility
2. You can also do `(flow one-flow another-flow)` and that is probably better. The only caveat is that a flow description is mandatory so you have to supply one, so it would be `(flow "my flow" one-flow another-flow)`. I didn't want to make it optional in order to encourage people to document the flows, the description acts as documentation and is also useful for generating the test failure descriptions.

That being said, the problems we were trying to solve with `wrap-with` can now be solved by supplying a custom runner here: https://github.com/nubank/state-flow/blob/master/src/state_flow/core.clj#L107

That is not the same as `wrap-with` macro that would enable for custom macros to be run for specific parts of the flow. But I'm very afraid that people would misuse this power if it becomes easy to do. WDYT?

philomates: What about running the flow directly via `:mfn` as a way to avoid calling `run!` inside of `wrap-with`?:

```clojure
(defmacro wrap-with
  [wrapper-fn flow]
  `(cats/do-let
    [world# (state/get)
     :let [pair# (~wrapper-fn ((:mfn ~flow) world#))]]
    (state/put (second pair#))
    (cats/return (first pair#))))

(defn put2 [w] (assoc w :value 2))
(defn put3 [w] (assoc w :value 3))

(defmacro wrapper
  [flow-fn]
  `(clojure.core/with-redefs [put2 put3]
     ~flow-fn))

(print (second (run (wrap-with wrapper (state/swap put2)) {})))
```

or if we want one specific for redefs, to avoid the need to define a wrapper macro, we could do
```clojure
(defmacro with-redefs
  [redefs flow]
  `(cats/do-let
    [world# (state/get)
     :let [pair# (clojure.core/with-redefs ~redefs
                   ((:mfn ~flow) world#))]]
    (state/put (second pair#))
    (cats/return (first pair#))))
```

I think that this would play well with whatever run function is defined

sovelten:
I'll be closing this as I think it has a lot of potential of being misused. We can always reopen if there is a good use case for it.

src/state_flow/core.clj
dchelimsky: > It might be a total hack but you could store the runner as metadata of the flow and have the `wrap-with` look up the runner of the flow arg and use it

We actually implemented this with a first-class step: https://github.com/nubank/state-flow/blob/5919a165ea568b9278dffeb73b44dc82c0dd7f3c/src/state_flow/core.clj#L121

PR #34, opened 2019-08-29
Allow str expr to be valid flow desc

philomates:
follow-up to https://github.com/nubank/state-flow/pull/31 which will allow for us to build better sub-flow description strings.

For example
```clojure
(defn my-subflow [arg1]
  (flow (str "my-subflow with arg" arg1) ...))
```

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -37,7 +37,7 @@
 
 (defn string-expr? [x]
   (or (string? x)
-      (and (sequential? x)
+      (and (list? x)
            (or (= (first x) 'str)
                (= (first x) 'clojure.core/str)))))
 
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -37,11 +37,17 @@
   (m/mlet [desc-list (state/gets #(-> % :meta :description))]
     (m/return (description->string desc-list))))
 
+(defn string-expr? [x]
+  (or (string? x)
+      (and (sequential? x)
+           (or (= (first x) 'str)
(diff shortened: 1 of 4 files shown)


src/state_flow/core.clj
```diff
@@ -35,11 +35,17 @@
...
+      (and (sequential? x)
+           (or (= (first x) 'str)
+               (= (first x) 'clojure.core/str)))))
+
 (defmacro flow
   "Defines a flow"
   {:style/indent :defn}
   [description & flows]
-  (when-not (string? description)
+  (when-not (string-expr? description)
```
achesnais: Curious, have you considered checking the string-ness at runtime, to allow for arbitrary string expressions?

src/state_flow/core.clj
```diff
@@ -35,11 +35,17 @@
   (m/mlet [desc-list (state/gets #(-> % :meta :description))]
     (m/return (description->string desc-list))))
 
+(defn string-expr? [x]
+  (or (string? x)
+      (and (sequential? x)
```
achesnais: Shouldn't this be `list`?

src/state_flow/core.clj
philomates: Yeah, I discussed it a little in https://github.com/nubank/state-flow/pull/31#pullrequestreview-272742394. I don't have any strong opinions, but I guess it would simplify this code.

ericvm: You can check string-ness at macro expansion time, just evaluate the expression and check it is a string. I think it would be cleaner

PR #35, opened 2019-08-29
Idea: show flow line numbers in failure message

philomates:
Since you can arbitrarily nest flows (which is amazing), you sometimes end up with test failures from nested flows for which it is difficult to know how far along the parent flows are.

If we could annotate each flow description with a line number and maybe file name, it would make it much quicker to track down where failures are coming from.

I tried to quickly hack a proof of concept for this in this PR.

Do others think something like this would be useful?

### Given

```clojure
(ns state-flow.lines-test
  (:require [cats.core :as m]
            [cats.data :as d]
            [cats.monad.state :as state]
            [midje.sweet :refer :all]
            [state-flow.core :as state-flow]
            [state-flow.state :as sf.state]))

(def increment-two-value
  (state/swap (fn [s] (update s :value #(+ 2 %)))))

(def failed-validate
  (state-flow/flow "inner validate"
    (state-flow/verify "value incremented by 3"
          (state/gets #(-> % :value)) 3)))

(def nested-flow
  (state-flow/flow "root"
    (state-flow/flow "child1" increment-two-value)
    (state-flow/flow "child2" increment-two-value)
    failed-validate
    (state-flow/verify "value incremented by 3"
      (state/gets #(-> % :value)) 3)))

(let [world {:value 0}]
  (state-flow/run nested-flow world))
```

### Before
```
FAIL root -> inner validate -> value incremented by 3 at (lines_test.clj:15)
Expected:
3
Actual:
4

FAIL root -> value incremented by 3 at (lines_test.clj:23)
Expected:
3
Actual:
4
```

### After
```
FAIL root -> inner validate (line 14) -> value incremented by 3 at (lines_test.clj:15)
Expected:
3
Actual:
4

FAIL root (line 22) -> value incremented by 3 at (lines_test.clj:23)
Expected:
3
Actual:
4
```

## Caveat

Ideally the
```
FAIL root -> inner validate (line 14) -> value incremented by 3 at (lines_test.clj:15)
```

would be
```
FAIL root (line 21) -> inner validate (line 14) -> value incremented by 3 at (lines_test.clj:15)
```

but this isn't possible in this approach because line metadata isn't present for variables (ie symbols at macro expansion time) so:
```clojure
(flow
  (something) ;; will have line-number metadata
  sub-flow-var ;; won't have line-number metadata :(
```
I don't really know why this is because symbols, just like lists, can have metadata associated with them.

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -29,13 +29,28 @@
      (update-in s [:meta :description] #(pop %)))))
 
 (defn description->string
-  [description]
-  (clojure.string/join " -> " description))
+  [descriptions lines]
+  (clojure.string/join " -> "
+                       (map (fn [desc line] (if (nil? line)
+                                              desc
+                                              (str desc "(" line ")")))
+                            descriptions
+                            lines)))
 
 (defn get-description
   []
-  (m/mlet [desc-list (state/gets #(-> % :meta :description))]
-    (m/return (description->string desc-list))))
+  (m/mlet [desc-list (state/gets #(-> % :meta :description))
+           lines-list (state/gets #(-> % :meta :lines))]
+    (m/return (description->string desc-list lines-list))))
+
+
+(defn- with-lines [flows]
+  (mapcat
+    (fn [flow]
+      [`(state/swap (fn [s#] (update-in s# [:meta :lines] (fn [l#] (update-description l# ~(-> flow meta :line))))))
+       flow
+       `(state/swap (fn [s#] (update-in s# [:meta :lines] (fn [l#] (pop l#)))))])
+    flows))
 
 (defmacro flow
   "Defines a flow"
@@ -47,7 +62,7 @@
                    '[(state/swap identity)])]
     `(m/do-let
        (push-meta ~description)
-       [ret# (m/do-let ~@flows')]
+       [ret# (m/do-let ~@(with-lines flows'))]
        pop-meta
        (m/return ret#))))
 


src/state_flow/core.clj
```diff
@@ -47,7 +62,7 @@
                    '[(state/swap identity)])]
     `(m/do-let
        (push-meta ~description)
-       [ret# (m/do-let ~@flows')]
+       [ret# (m/do-let ~@(with-lines flows'))]
```
ericvm: Instead of `with-lines`, I would add meta information to `push-meta` and `pop-meta`.
Also, since this is internal implementation detail there is no backward compatibility you need to maintain.

I was thinking about something along this:
```clojure
(defn push-meta
  [meta description]
  (fn [s] (update-in s [::meta :flows] #(update-flow-metadata % meta description))
```

Also moving to qualified keyword `::meta` is something that @caioaao suggested a while ago. WDYT?

philomates: I think to get line info we have to add a new step before each flow step so that we can read the flow's syntax metadata. Hence, we can't also get the description at the same time because that is only available inside the flow step.

ericvm: You can use &form to grap metadata for current macro (https://clojure.org/reference/macros#_special_variables)
```clojure
 (let [flow-meta (meta &form]
    `(m/do-let
       (push-meta ~flow-meta ~description)
    ...
```

something like this

dchelimsky:
@philomates I implemented this on the `dc/line-number-on-steps` branch. Please have a look at https://github.com/nubank/state-flow/compare/dc/line-number-on-steps and let me know what you think. If you like it, I can push that to this branch so all the PR convo is in one place.

philomates:
addressed in https://github.com/nubank/state-flow/pull/51

PR #41, opened 2019-12-10
Make state-flow.state/swap work more like Clojure's swap

dchelimsky:
Many of Clojure's higher-order functions like update, update-in, swap!, etc, take varargs in order to support passing a function and it's arguments as first class arguments rather than wrapping it all up in an anonymous function e.g.

``` clojure
(update {:a 1} :a + 2)

;; vs

(update {:a 1} :a #(+ % 2))
```

This PR makes `state-flow.state/swap` work the same way. This is backward compatible: you can still provide a single function argument, but you can choose to use the more idiomatic approach.

``` clojure
;; both of these work

(def inc-value (state/swap update :value inc)))

(def inc-value (state/swap #(update-in % [:value] inc)))
```

### DIFF
--- a/src/state_flow/state.clj
+++ b/src/state_flow/state.clj
@@ -68,8 +68,8 @@
   (state/put s error-context))
 
 (defn swap
-  [f]
-  (state/swap f error-context))
+  [f & args]
+  (state/swap (fn [s] (apply f s args)) error-context))
 
 (defn wrap-fn
   "Wraps a (possibly side-effecting) function to a state monad"
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ fetches the `<value>` inside `<number>`:
 For updating the state we can use `state/swap`. If we want to write a flow that will increment value by one, it could be done like this:
 
 ```clojure
-(def inc-value (state/swap #(update-in % [:value] inc)))
+(def inc-value (state/swap update :value inc)))
 (state-flow/run! inc-value {:value 4})
 ; => [{:value 4} {:value 5}]
 ```
--- a/samples/tutorial.clj
+++ b/samples/tutorial.clj
@@ -21,7 +21,7 @@ Runner
 (state-flow/run! get-value {:value 4})
 ; => [4 {:value 4}]
 
-(def inc-value (state/swap #(update-in % [:value] inc)))
+(def inc-value (state/swap update :value inc))
 (state-flow/run! inc-value {:value 4})
 ; => [{:value 4} {:value 5}]
 
@@ -100,5 +100,3 @@ Asynchronous tests
 
 (state-flow/run! with-async-success {:value (atom 4)})
 ;=> [5 {:value (atom 5)}]
-
-
--- a/test/state_flow/core_test.clj
+++ b/test/state_flow/core_test.clj
@@ -41,7 +41,7 @@
   (state-flow/flow "root"
     [original (state/gets :value)
      :let [doubled (* 2 original)]]
-    (sf.state/swap #(assoc % :value doubled))
+    (sf.state/swap assoc :value doubled)
     (midje/verify "value is doubled"
       (state/gets #(-> % :value)) doubled)))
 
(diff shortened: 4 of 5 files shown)


src/state_flow/state.clj
```diff
@@ -68,8 +68,8 @@
   (state/put s error-context))
 
 (defn swap
-  [f]
-  (state/swap f error-context))
+  [f & args]
+  (state/swap (fn [s] (apply f s args)) error-context))
```
sovelten: I must admit I'm not a fan of this style in Clojure, but it is more idiomatic. This might break compatibility with https://github.com/nubank/cats/blob/master/src/cats/monad/state.cljc#L133, but I don't see any problem with adding the option, it makes sense to me. Maybe adding there is also a good idea.

Can you add a message in the changelog and increase the library version?

dchelimsky: > Can you add a message in the changelog and increase the library version?

Coming shortly.

I want to understand your compatibility concern before I merge this. Can you elaborate?

dchelimsky: I also want to understand what you dislike about the approach/style.

sovelten: I think I was wrong about compatibility, sorry. Because we are already calling `swap` from cats directly, so not really an issue. Either way, we can also change `swap` on `cats` to support this approach.

About the style, it's just personal taste. It looks less readable to me, it's not immediately clear that the following arguments are going to be applied to the function it was supplied. Also, once it is variadic it is harder to extend by adding more arguments. If for some reason you have to pass two functions as arguments it is not possible to both in this way. I'd much prefer if clojure had auto-currying or we just pass `(partial f & args)`, arguably more consistent. WDYT?

dchelimsky: I understand your position in general. Believe it or not, I'm coming from a place of consistency as well. The fact that `swap` sounds like `swap!`, and expresses a similar metaphor (change a value in some wrapped state/atom), means that it would be reasonable for a Clojure developer to assume that they work similarly. We should probably also consider consider renaming (or aliasing) the `put` function to `reset` because `put` generally means "add to a collection" in other contexts like http, and `reset` makes a good pair with `swap` (i.e. `swap!` and `reset!` on atoms). WDYT?

sovelten: I agree with both changes.
Just a minor consideration for `put/reset`, it's not a good practice to use it inside a flow, because you'll lose meaningful execution metadata that is stored in the state.

sovelten: Actually, it turns out there are some use cases for `put/reset`. This code in tombamento-financeira https://github.com/nubank/tombamento-financeira/blob/master/src/acceptance_tests/system.clj#L42 could be rewritten with it:
```clojure
(defn with-service [service flow]
  (state-flow/flow "Run flow with a specific service"
    [old-world (state-flow.state/swap (fn [world] (assoc world :system (-> world :systems service))))
     ret flow]
    (state/reset old-world)
    (state/return ret)))
```

More concise and more readable.

dchelimsky: OK, so while we're on a roll here, how about `s/gets/apply` as well, e.g.

```clojure
(s/apply inc)

;; and/or

(s/apply + 2)
```

sovelten: `gets`, `get`, `put` and `modify` are consistent with the state monad in Haskell https://wiki.haskell.org/State_Monad

The original author of the state monad in https://github.com/funcool/cats used `swap` instead of `modify` to be more familiar to the clojure programmer, so I don't dislike the idea of renaming everything instead of building a Frankenstein with some names taken from Haskell and other names from Clojure.

My concern here for `apply` is: does it transmit the idea that the state is not supposed to change? At least not directly.

Because `(state/gets f)` ends up building this sort of function: `(fn [state] [(f state) state])`, so the function is only applied to the return value, not to the state, like `swap`.

Going forward in this line of thinking, what would be the equivalent of `state/get`? Maybe something like `state/deref`?

dchelimsky: This is really interesting. My initial knee-jerk reaction was to make this all feel more like Clojure, but the fact that we're dealing with pairs returned by a function whose execution is delayed makes me think we should clarify that's what's happening without getting into the theory behind monads. We can do that by renaming `swap` to `modify`, referencing the Haskell docs for those who want to learn more in the README, coupled with a very high level explanation of what these functions return (i.e. think of them as returning functions that return pairs).

WDYT?

dchelimsky: Alternatively, if we want to stick to the path of making it more like clojure names, we could add "!" to the functions that modify state, so `deref`, `apply`, `swap!`, and `reset!`. This might resolve the concern you raised about understanding the semantics of `apply`.

dchelimsky: One thing I notice, btw, is that cats' `put` doesn't work like the `put` described here: https://wiki.haskell.org/State_Monad (which returns what looks to my naive eyes like an empty list which I'm guessing from other functional languages is the same as null). Would changing our implementation to return `(State. (fn [s] [nil, new-value]))` break down?

dchelimsky:
Went with the approach in #42 instead.

PR #42, opened 2019-12-17
use Haskell's names for State monad functions

dchelimsky:
This PR deprecates the `swap` function, which sounds like a Clojure core function, but works differently, and replaces it with the `modify` function, aligning all of the state primitives with the language of Haskell's State Monad: https://wiki.haskell.org/State_Monad

It also `defn`s the `gets` and `return` functions to support documentation than supported by just passing them through a `def`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 An integration testing framework for building and composing test flows with support for clojure.test and midje
 
-The StateFlow library aim to provide a compositional way in which one could implement integration tests. The main advantage of this approach is to reduce coupling between test steps and allow for more reusability and composability of flows.
+StateFlow provides a compositional approach to implementing integration tests. The goal is to reduce coupling between test steps in order to support reuse and composition of flows.
 
 ## The flow macro
 
-Defining a flow is done by using the `flow` macro, which expects a description as first parameter and a variable number of steps that can be other flows, bindings or primitives.
+Defining a flow is done with the `flow` macro, which expects a description and a variable number of steps that can be other flows, bindings or primitives.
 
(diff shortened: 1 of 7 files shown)


dchelimsky:
@sovelten what do you think of this rather than #41?

README.md
```diff
@@ -27,73 +27,81 @@ a system using [Stuart Sierra's Component](https://github.com/stuartsierra/compo
 
 ### Primitives
 
-The primitives are the fundamental building blocks of a flow and are enough to build
-any kinds of flow.
-Below we list the main primitives and the kind of function they represent:
+Primitives are the fundamental building blocks of flows and are
+enough to build any kind of flow. Each one returns a function of the
+state. These functions are wrapped in Records in order to support
+background processing, but you can just think of them as functions.
```
sovelten: in order to support implementation of protocols maybe? not clear to me what background processing means

src/state_flow/state.clj
```diff
@@ -60,17 +60,35 @@
...
-  (state/put s error-context))
+  "Returns the equivalent of (fn [state] [state, new-state])"
+  [new-state]
+  (state/put new-state error-context))
 
-(defn swap
+(defn modify
+  "Returns the equivalent of (fn [state] [state, (swap! state f)])"
   [f]
   (state/swap f error-context))
```
dchelimsky: This can change later if we add `modify` to cats, with no penalty for users of StateFlow after this PR.

sovelten:
Probably could change mentions of `state/swap` to `state/modify` on the tests as well

PR #45, opened 2019-12-17
Add `testing` macro to cljtest

caioaao:
It looks like it doesn't break anything (amount of fails when running the test is the same as before the PR), but the test output was confusing, so I'm not sure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [2.0.5]
+- Add `state-flow.labs` namespace for experimental features
+- Add `state-flow.labs.cljtest/testing`
+
 ## [2.0.4]
 - update `times-to-try` default from `100` to `5` and `sleep-time` default from `10` to `200`
 
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject nubank/state-flow "2.0.4"
+(defproject nubank/state-flow "2.0.5"
   :description "Postman-like integration testing with composable flows"
   :url "https://github.com/nubank/state-flow"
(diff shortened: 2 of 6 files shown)


sovelten:
You need to make sure it actually works with `lein test` or equivalent because sometimes it prints on the REPL that the assertions passed/failed but it doesn't actually register the assertion.

I remember trying to use `testing` before with no success, but I don't remember if it was for registering the whole flow or just a single test.

I think we can iterate on this, though. What do you think are the main advantages of using `testing`?

The way it is right it does not support test probing and it doesn't support passing a flow as an argument.

caioaao:
`lein test` output (after the exceptions):
```
Ran 1 tests containing 2 assertions.
0 failures, 0 errors.
```
Running `lein midje` I get this:
```
>>> Midje summary:
FAILURE: 3 checks failed.  (But 55 succeeded.)
```
But the 3 checks are the same as in master apparently.

About the advantages of using `testing`: basically just being able to reuse existing stuff without having to reimplement it here. One example (and the reason I wrote it atm) is `thrown-match?` from matcher-combinators, but I feel the flexibility of having a `testing` block inside the flow outweights the benefits. also, probing and other stuff imo can be done inside the assertions instead of some automatic thing

sovelten:
What do you think about putting this on some labs namespace (like cats does https://github.com/nubank/cats/tree/master/src/cats/labs)?

Than we can play around with this approach and other similar experiments and evolve it without fear of breaking compatibility or supporting legacy stuff.

caioaao:
That's a great idea. I'll move the stuff

sovelten:
just add CHANGELOG and bump the version ;)

PR #49, opened 2019-12-26
Convert tests to use clojure.test

dchelimsky:
This PR updates all tests to use clojure.test, and also introduces a run-flow! macro to support testing flows that include failing tests without reporting those failures (but still being able to make assertions about the expected failures).

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -50,7 +50,7 @@
   (when-not (string-expr? description)
      (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
   (let [flows' (or flows
-                   '[(m/return nil)])]
+                   `[(m/return nil)])]
     `(m/do-let
       (push-meta ~description)
       [ret# (m/do-let ~@flows')]
@@ -64,7 +64,7 @@
   (state/run flow initial-state))
 
 (defn run!
-  "Like run, but prints a log and throws error when flow fails with an exception"
+  "Like run, but prints a log and throws an error when the flow fails with an exception"
   [flow initial-state]
   (let [result (run flow initial-state)]
     (when (e/failure? (first result))
--- a/test/state_flow/cljtest_test.clj
+++ b/test/state_flow/cljtest_test.clj
@@ -142,6 +142,3 @@
   (is (match? {:value 1
                :map   {:a 1 :b 2}}
               (second ((:test (meta #'my-flow)))))))
-
-(comment
-  (t/run-tests))
--- a/test/state_flow/cljtest_test.clj
+++ b/test/state_flow/cljtest_test.clj
@@ -36,27 +36,27 @@
           (is (= {:a 2 :b 5} (:value state)))))
 
   (testing "failure case"
-    (let [{:keys [flow-res flow-state]}
+    (let [{:keys [flow-ret flow-state]}
           (th/run-flow (cljtest/match? "contains with monadic left value"
                                                       (state/gets :value)
                                                       (matchers/equals {:a 1 :b 5}))
                     {:value {:a 2 :b 5}})]
-      (is (= {:a 2 :b 5} flow-res))
+      (is (= {:a 2 :b 5} flow-ret))
       (is (= {:a 2 :b 5} (:value flow-state)))))
 
   (testing "add two with small delay"
     (let [state  {:value (atom 0)}
-          {:keys [flow-res]}
+          {:keys [flow-ret]}
           (th/run-flow
             (flow ""
               (th/delayed-add-two 100)
               (cljtest/match? "" get-value-state 2))
             state)]
-      (is (= 2 flow-res))))
+      (is (= 2 flow-ret))))
 
   (testing "we can tweak timeout and times to try"
     (let [state  {:value (atom 0)}
-          {:keys [report-data flow-res flow-state]}
+          {:keys [report-data flow-ret flow-state]}
           (th/run-flow
            (flow ""
              (th/delayed-add-two 100)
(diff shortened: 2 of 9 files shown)


test/state_flow/state_test.clj
```diff
@@ -1,38 +1,45 @@
...
+                                      (m/return (midje/fact "z is equal to initial state incremented by two"
+                                                            z => (+ x 2))))
+        state-with-assertion  (m/mlet [x increment-state
+                                       y increment-state
+                                       z (sf.state/get)]
+                                      (m/return (is (= z (+ x 2)))))]
+    (testing "modify state with get and put"
+      (is (= (d/pair 2 3)
+             (state/run increment-state 2))))
+    (testing "modify state with modify"
```
dchelimsky: This one is new. Added it to contrast with the example before.

test/state_flow/state_test.clj
```diff
@@ -1,38 +1,45 @@
...
-            [cats.protocols :as p]
-            [midje.sweet :refer :all]
+            [midje.sweet :as midje]
             [state-flow.state :as sf.state]))
 
-(def postincrement
-  (m/mlet [x (sf.state/get)
-           _ (sf.state/put (+ x 1))]
-    (m/return x)))
+(deftest test-state
```
sovelten: What do you think about not using `test` as part of the name? Since `deftest` already implies it is a test. I have no clear sides on this, but prefer on avoiding redundancy.

test/state_flow/state_test.clj
```diff
@@ -1,38 +1,45 @@
...
+  (let [increment-state       (m/mlet [x (sf.state/get)
+                                       _ (sf.state/put (inc x))]
+                                      (m/return x))
+        double-state          (sf.state/modify #(* 2 %))
+        state-with-fact-check (m/mlet [x increment-state
+                                       y increment-state
+                                       z (sf.state/get)]
+                                      (m/return (midje/fact "z is equal to initial state incremented by two"
```
sovelten: Since you are refactoring tests, midje related tests should be on midje namespace, same for clojure.test.

So maybe these tests can just go away, they are also being tested there.

dchelimsky: Agreed. Done.

test/state_flow/state_test.clj
dchelimsky: https://github.com/bbatsov/clojure-style-guide suggests names should be xxx-test vs test-xxx. I prefer test-xxx because that's a verb form. Either way, the convention is to include "test" in the name. Keep in mind that `deftest` is a macro which produces a function. A function named "test-state" (or "state-test") makes more sense than a function named "state" :)

dchelimsky: Oh, well, then again, https://clojure.github.io/clojure/clojure.test-api.html suggests to leave test out. Flip a coin?

sovelten: The test namespace is supposed to contain tests, one possible convention to adopt is that the tests are called the same way as the functions they are testing. I honestly have used both approaches at different places. I tend to favor leaving test out. Stuart Sierra recommendations are often wise.

test/state_flow/test_helpers.clj
```diff
@@ -2,20 +2,41 @@
...
-  (state/state (fn [world]
+  "Changes state in the future"
+  (state/state (fn [state]
                  (future (do (Thread/sleep delay-ms)
-                             (swap! (:value world) + 2)))
-                 (d/pair nil world))))
+                             (swap! (:value state) + 2)))
+                 (d/pair nil state))))
 
+(defmacro run-flow
```
dchelimsky: @sovelten @philomates this allows us to assert test failures or exceptions in flows without the expected exceptions and failures showing up in stdout or stderr or the clojure.test report. Thoughts?

src/state_flow/core.clj
```diff
@@ -50,7 +50,7 @@
   (when-not (string-expr? description)
      (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
   (let [flows' (or flows
-                   '[(m/return nil)])]
+                   `[(m/return nil)])]
```
dchelimsky: FYI: this caused an error when an empty flow was run in a namespace that did not require `cats.core :as m`.

test/state_flow/state_test.clj
mpivaa: I'm adding `test` even on flows. It makes easier for editors and test runners.

test/state_flow/cljtest_test.clj
```diff
@@ -1,10 +1,10 @@
 (ns state-flow.cljtest-test
   (:require [clojure.test :as t :refer [deftest testing is]]
+            [matcher-combinators.test] ;; loads match? assertions
```
philomates: sometimes we write `[matcher-combinators.test :refer [match?]]` to help linters out

test/state_flow/midje_test.clj
```diff
@@ -1,38 +1,42 @@
...
-      (state/run (midje/verify-probe "just with monadic left value"
-                                     (sf.state/get) (just {:a 2 :b 5}) {}) val) => (d/pair val val))))
+      (= (d/pair val val)
+         (state/run (midje/verify-probe "contains with monadic left value"
+                                        (sf.state/get) (contains {:a 2}) {}) val)
+         (state/run (midje/verify-probe "just with monadic left value"
+                                        (sf.state/get) (just {:a 2 :b 5}) {}) val)))))
+
+(comment
+  (t/run-tests))
```
philomates: :scissors: :)

dchelimsky: Done.

test/state_flow/cljtest_test.clj
dchelimsky: Done.

test/state_flow/state_test.clj
dchelimsky: I left it in a couple of places where it would otherwise clobber functions we're referring from other namespaces.

PR #50, opened 2020-01-07
move flow description to meta of the State object

dchelimsky:
- reduces noise in, especially in docs
- also test description via get-description api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [2.1.1]
+- Moved flow descriptions to the State object's meta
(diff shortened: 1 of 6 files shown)


dchelimsky:
Thanks @philomates. I added another commit to clean up a couple of implementation details. Still approve?

philomates:
yup, even cleaner
Even though this is sorta an internal thing, I think adding it to the changelog is useful in case any users have built systems on top of `state-flow`

dchelimsky:
Some day I'll remember to update the changelog before you bring it up.

PR #52, opened 2020-01-08
get description from meta (not :meta) in run! fn

dchelimsky:
Also add `top-level-description` function and refactor some internals

- simplify description functionality
  - no need to maintain log
- document some non-API fns that must be public as "for internal use"

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -1,45 +1,70 @@
 (ns state-flow.core
   (:refer-clojure :exclude [run!])
-  (:require [cats.context :as ctx]
+  (:require [clojure.string :as str]
             [cats.core :as m]
             [cats.monad.exception :as e]
             [state-flow.state :as state]
             [taoensso.timbre :as log]))
 
-(defn push-description
-  [description-log new-description]
-  (if (nil? description-log)
-    [[new-description]]
-    (conj description-log (conj (last description-log) new-description))))
+;; From time to time we see the following error when trying to pretty-print
+;; Failure records:
+;;
+;;   Unhandled java.lang.IllegalArgumentException
+;;   Multiple methods in multimethod 'simple-dispatch' match dispatch
+;;   value: class cats.monad.exception.Failure -> interface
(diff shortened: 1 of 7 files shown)


dchelimsky:
@Andre0991 this might impact visual-flow. Let me know if you have any questions.

src/state_flow/core.clj
```diff
@@ -1,45 +1,70 @@
...
+(defmethod clojure.pprint/simple-dispatch cats.monad.exception.Failure [f]
+  (pr f))
 
-(defn pop-description
-  [description-log]
-  (conj description-log (pop (last description-log))))
-
-(defn update-meta [s k & args]
-  (with-meta s (apply update (meta s) k args)))
+(defn ^:private alter-meta!*
```
drayah: Why use `^:private` here for this `defn` but in other parts of the namespace `defn-`?

dchelimsky: No reason. Looking around at other namespaces, `^:private` is used everywhere else, so I went with that for all the private things in this namespace, too.

src/state_flow/core.clj
```diff
@@ -1,58 +1,85 @@
...
    (fn [s]
-     (update-meta s :description push-description description))))
+     (-> s
+         (alter-meta!* update :description-stack (fnil conj []) description)
+         (alter-meta!* update :top-level-description #(or % description))))))
 
 (def pop-meta
+  "Returns a flow that will modify the state metadata.
+
+  For internal use. Subject to change."
```
Andre0991: Not related to this PR, just some random thought: It would be nice if there were a clearer separation between public API and internal stuff. There's this "Public API" comment, but we still need to add this sort of warning in the docstrings, like above.

dchelimsky: Ideally we'd make `pop-meta` private (it is, conceptually), but we ran into some problems related to macro expansion of these macros in other namespaces. Probably fixable, but we just didn't narrow it down yet.

PR #53, opened 2020-01-08
add :let binding to README

diegospd:


README.md
```diff
@@ -79,6 +79,12 @@ Bindings take advantage of the return values of flows to compose other flows and
 `[(<symbol> <flow/primitive>)+]`
 
 They work pretty much like `let` bindings but the left symbol binds to the _return value_ of the flow on the right.
+It's also possible to do non-flow bindings inside the same vector using the `:let` keyword:
+
+```clojure
+[(<symbol> <flow/primitive>)
+ :let [<symbol> <non-monadic value>]]
```
philomates: maybe `value` -> `expression`

philomates: and also `non-monadic` -> `non-flow` given that we are trying to avoid mentions of monads :)

PR #55, opened 2020-01-09
fix failing test

dchelimsky:
For reasons unknown, this test was passing when running `lein test` but threw an exception when running in an REPL (`=` requires 2 args).


dchelimsky:
Given that this fixes a broken test and nothing else, I'm going ahead and merging.

PR #62, opened 2020-01-16
Add arity with cleanup on `run!`

### LINKED ISSUE #59: cleanup should run even if exception occurs
In cljtest I expected the `cleanup` function to run even if an exception occurs.

caioaao:
Fixes #59

Also adds the original exception from the flow as the cause for the thrown exception so we don't depend on logging to find out what happened


caioaao:
This probably changes the library's behavior under some circumstances, so I'll map the differences out to make sure it's not breaking any use cases.

src/state_flow/core.clj
```diff
@@ -110,12 +110,18 @@
   ([flow]
    (run! flow {}))
   ([flow initial-state]
+   (run! flow initial-state identity))
+  ([flow initial-state cleanup]
```
dchelimsky: Any reason not to support running `cleanup` in `run` as well?

caioaao: Because it wouldn't make much of a difference, since `run` is not the default runner for `run*` and it doesn't throw, so the `cleanup` will be called inside `run*`. IMO it's still a hacky solution to avoid breaking changes though

dchelimsky: `run*` and `run!` both end up going through `run`. In my view, the only difference between `run` and run!` is that `run!` logs and throws failures, whereas `run` just returns them. In a test that uses `run` directly, we'd still want the ability to supply a `cleanup` function in the case of failures, no?

caioaao: In this case we only have `cleanup` on `run*`. I'd argue that `run` and `run!` don't care about what happens to the result or how to generate the initial state, that's why they don't have these options. But, through that reasoning `run!` doesn't make much sense either, which I would be in favor of deprecating if we have the ability to log and throw inside `run*`, like the `on-error` option I just pushed.

dchelimsky: I guess I'm thrown by the fact that cleanup is exposed as a argument to `run!`, but from what you're saying, your intent is only to ever use it in the context of `run*`.

In that case, I'd recommend moving the cleanup logic to `run*` and leaving `run!` alone. That would leave the APIs for `run` and `run!` aligned, and would also provide the opportunity to guarantee that `cleanup`happens even when you pass in a custom `runner` to `run*`. Right now, it only happens if you do pass in cleanup and don't pass in a runner.

caioaao: yup! that's what I meant. And it's already done :) (https://github.com/nubank/state-flow/pull/62/commits/c4ecdbf2c67d26c68956ce4256fa50d352a849d2)

caioaao: With all said, I'm also considering adding the `:deprecated` flag to `run!`, otherwise users will call it in a custom runner and the cleanup won't run

src/state_flow/core.clj
```diff
@@ -105,35 +105,49 @@
...
-  `init`, a function with no arguments that returns the initial state.
+  `init`, a function with no arguments that returns the initial state
   `cleanup`, function receiving the final state to perform cleanup if necessary
-  `runner`, function that will receive a flow and an initial state and execute the flow"
-  [{:keys [init cleanup runner]
-    :or   {init    (constantly {})
-           cleanup identity
-           runner  run!}}
+  `runner`, function that will receive a flow and an initial state and execute the flow
+  `on-error`, behavior when a flow fails with an exception. Options are `:log-and-throw` and `:ignore`"
```
dchelimsky: how about:

```
`on-error`, keyword indicating behavior when a flow fails with an exception. Options are `:log-and-throw` and `:ignore`
```

src/state_flow/core.clj
```diff
@@ -105,35 +105,49 @@
...
-        [_ final-state :as result] (runner flow initial-state)]
-    (cleanup final-state)
-    result))
+  (let [initial-state (init)
+        pair          (runner flow initial-state)]
+    (try
+      (cleanup (second pair))
+      (catch java.lang.Throwable t
+        (log/error t "Error when trying to cleanup after exception")))
+    (result-or-run-policy! pair on-error)))
```
dchelimsky: I find this a little confusing. WDYT about something like:

```clojure
(run-policy-on-error! pair on-error)
pair
```

Then `run-policy-on-error!` doesn't need to worry about returning the pair, and I think it's easier to grok from here what's going on without going and looking at `run-policy-on-error!`

Alternatives:

```clojure
(if (should-log-and-throw? pair on-error)
  (log-and-throw pair)
  pair)

;; or

(or (log-and-throw pair on-error) pair)
```

I understand that means a small bit of duplication, but I think any one of these is more clear than a function with "or" in its name.

src/state_flow/core.clj
dchelimsky: Great.

dchelimsky: Can we address whether to deprecate `run!` in a separate PR/convo?

src/state_flow/core.clj
caioaao: Totally agree. I changed this back and forth 30 times and wasn't happy with the result either hahah. I'd prefer maintaining the explicit conditional on the options though, so we can extend it easier for other unforeseen cases (just log would be an example) and so it breaks if an invalid option is passed

dchelimsky:
Thanks @caioaao. This looks good to me now. I could approve it but it doesn't require approval and I'd prefer to see feedback from @sovelten and/or @philomates before merging.

src/state_flow/core.clj
```diff
@@ -105,35 +105,52 @@
...
-    :or   {init    (constantly {})
-           cleanup identity
-           runner  run!}}
+  `runner`, function that will receive a flow and an initial state and execute the flow
+  `on-error`, keyword indicating behavior when a flow fails with an exception. Options are `:log-and-throw` and `:ignore`"
+  [{:keys [init cleanup runner on-error]
+    :or   {init     (constantly {})
+           cleanup  identity
+           runner   run
+           on-error :log-and-throw}}
```
philomates: If I understand this correctly, the behavior that is tuned via `on-error` can be encoded by passing in a different `runner`. 

For instance, `on-error :ignore` could be like using `:runner run` or `on-error :log-and-throw` is the same as `:runner run!`, right?

I imagine the idea came about to be able to ensure that `cleanup` is called. Would it be possible to make `cleanup` like a `finally` block (which is a breaking change but I see it as a bugfix breaking, so no need to deprecate things, in my opinion). The `cleanup` can then be written to skip specific logic if it gets an error value, which might be something you were trying to encode with `on-error`.

(I hope I'm not too far off the track with these suggestions, always easier to balance trade-offs when you've encountered these issues first hand)

caioaao: You can't use cleanup as a finally block since the exception is already thrown and we lose the result pair (so we can't call `cleanup` on the state). About the different runner: yeah, that was the previous idea, but it doesn't look as good and we tie the runner logic to the cleanup, which imo is a bad idea

caioaao:
@sovelten @philomates is it possible to unstuck this? This PR was a big improvement on my experience with the project and a boost to my productivity when using it btw :grimacing:

src/state_flow/core.clj
```diff
@@ -105,35 +105,52 @@
...
+(defn- log-and-throw!
+  [pair]
+  (let [description (->> pair second description-stack format-description)
+        message     (str "Flow " "\"" description "\"" " failed with exception")]
+    (log/info (m/extract (first pair)) message)
+    (throw (ex-info message {} (m/extract (first pair))))))
+
+(defn- run-policy-on-error!
+  "If flow fails with an exception, runs the supplied error policy"
+  [pair on-error]
```
dchelimsky: Names like `on-error` are often used to pass functions in, but we're passing in a key representing a policy. How about something like:

```clojure
(defn- run-policy-on-error!
  "Runs the policy named by policy-key iff pair represents a flow failure"
  [pair policy-key]
...
```

src/state_flow/core.clj
philomates: what if we changed the implementation of the error short-circuiting thing to also return the state? Then cleanup could do its thing regardless.

src/state_flow/core.clj
```diff
@@ -105,35 +105,52 @@
...
-  `init`, a function with no arguments that returns the initial state.
+  `init`, a function with no arguments that returns the initial state
   `cleanup`, function receiving the final state to perform cleanup if necessary
-  `runner`, function that will receive a flow and an initial state and execute the flow"
-  [{:keys [init cleanup runner]
-    :or   {init    (constantly {})
-           cleanup identity
-           runner  run!}}
+  `runner`, function that will receive a flow and an initial state and execute the flow
+  `on-error`, keyword indicating behavior when a flow fails with an exception. Options are `:log-and-throw` and `:ignore`"
```
dchelimsky: Per my other comment, `on-error` sounds (to me) like a function. How about e.g. `error-policy` or `error-policy-key`?

src/state_flow/core.clj
caioaao: Agreed, but instead I changed it to receive a function, which might be more in-line with the API and more flexible

dchelimsky: Or actually pass in functions :)

Thanks!

src/state_flow/core.clj
caioaao: we can't, since if it returns the state, then it doesn't short circuit :) We could put it in the `ex-data`, but I think that'll pollute the output.

src/state_flow/core.clj
sovelten: I agree with deprecating `run!` in favor of an error handler option. I think it makes things clearer and less coupled.

Also, details aside, the general idea of the pull request makes sense 👍

dchelimsky: @sovelten do you think that (deprecating `run!`) should happen in this PR?

philomates: I think if new behaviors are introduced that make `run!` not preferred/redundant, then we should also deprecate it in the same release

src/state_flow/core.clj
```diff
@@ -105,35 +105,56 @@
...
-    (cleanup final-state)
-    result))
+  (let [initial-state (init)
+        pair          (runner flow initial-state)]
+    (try
+      (cleanup (second pair))
+      (catch java.lang.Throwable t
+        (log/error t "Error when trying to cleanup after exception")))
+    (or (run-policy-on-error! pair on-error)
+        pair)))
```
philomates: if `on-error` returns `nil` then `pair` will be returned, which might be considered unexpected

src/state_flow/core.clj
```diff
@@ -105,35 +105,56 @@
...
+           on-error log-and-throw-error!}}
    flow]
-  (let [initial-state              (init)
-        [_ final-state :as result] (runner flow initial-state)]
-    (cleanup final-state)
-    result))
+  (let [initial-state (init)
+        pair          (runner flow initial-state)]
+    (try
+      (cleanup (second pair))
```
philomates: I wonder if we want to run `cleanup` even when errors occur? That is what is happening right now, right?
Doing so would mean that `cleanup` needs to check for errors as well

caioaao: I think running cleanup when errors occur is the most important usage for the cleanup tbh, otherwise we could just append a step on every flow that did the cleanup. Without it running multiple tests or REPL development is intractable. Consider the basic example of testing an HTTP server. Our `:init` would do something like `(start-server :port 9000)`. If we don't cleanup on exceptions, we can't re-run the tests since we lost the server reference when state was lost.

About `cleanup` checking for errors: I don't think it should care about errors. `cleanup` should just _make sure_ every resource is freed after a flow.

src/state_flow/core.clj
caioaao: It's a new behavior, so it won't be a breaking change anyways.

About being considered unexpected: that's true, but I also think throwing information away is not a good practice. And if an exception is ignored, I don't see a reason not to return the result.

src/state_flow/core.clj
```diff
@@ -105,35 +105,56 @@
                (second pair))
        pair))))
 
+(defn log-and-throw-error!
+  "Error handler that logs the error and throws an exception to notify the flow
+  has failed."
+  [pair]
+  (let [description (->> pair second description-stack format-description)
```
sovelten: I think you need to use `top-level-description` function instead of `description-stack`, no?

src/state_flow/core.clj
```diff
@@ -105,35 +105,56 @@
...
+        message     (str "Flow " "\"" description "\"" " failed with exception")]
+    (log/info (m/extract (first pair)) message)
+    (throw (ex-info message {} (m/extract (first pair))))))
+
+(defn ignore-error
+  "No-op error handler that simply ignores the error."
+  [pair])
```
sovelten: `(def ignore-error identity)`?

src/state_flow/core.clj
dchelimsky: The PR is showing conflicts - this is one of them.

src/state_flow/core.clj
```diff
@@ -105,35 +105,56 @@
...
+
+(defn- run-policy-on-error!
+  "If flow fails with an exception, runs the supplied error policy"
+  [pair on-error]
+  (when (e/failure? (first pair))
```
sovelten: What would happen when there is an exception caused by the runner? That is a possibility, specially with custom runners.

test/state_flow/core_test.clj
```diff
@@ -96,6 +97,21 @@
...
+                                              :cleanup (fn [& _] (swap! cleanup-runs inc))}
+                              bogus-flow)))
+      (is (= 1 @cleanup-runs))))
+
+  (testing "flow with cleanup and exception, but ignoring it instead"
+    (let [result (state-flow/run* {:init     (constantly {:value 0})
+                                   :on-error state-flow/ignore-error}
+                   bogus-flow)]
+      (is (e/exception? (first result)))
+      (is (match? {:value 2} (second result)))))
```
sovelten: There should be a test for:
* Exception on flow :heavy_check_mark: 
* Exception within cleanup
* Exception within runner

dchelimsky: We also need to be clear about runtime exceptions vs assertion failures. The implications of a runtime exception early on in a flow is different from a failed assertion IMO, though we could decide to treat them all the same.

src/state_flow/core.clj
caioaao: I wanted to add the documentation and arglist here. btw, is there a reason to return the pair?

src/state_flow/core.clj
caioaao: I'll have to think about that. What do you think should be the expected behavior?

My guess is, since this only runs on failures reported on the result pair, exceptions outside the flow itself would (and should) be handled externally as well.

test/state_flow/core_test.clj
caioaao: @dchelimsky you might want to weigh in on this then: https://github.com/nubank/state-flow/issues/21 :)

src/state_flow/core.clj
sovelten: I'm ok with that assumption, but there is a try/catch for exception on cleanup, following the same idea, it makes sense to have it for exceptions on the runner. Just in the spirit of providing a better failure message. WDYT?

src/state_flow/core.clj
sovelten: I would rather think "is there a reason not to return the pair?"
But this is just :nail_care: anyway.

src/state_flow/core.clj
dchelimsky: > exceptions outside the flow itself would (and should) be handled externally as well.

If we're going to catch them, then why not add them to the pair as well? If we're treat them as something different, we could also choose to not catch them.

test/state_flow/core_test.clj
dchelimsky: Às ordens: https://github.com/nubank/state-flow/issues/21#issuecomment-582001061

src/state_flow/core.clj
caioaao: > ...but there is a try/catch for exception on cleanup...

Actually this was left from the old code and can be changed to a `try ... finally` block.

> If we're going to catch them, then why not add them to the pair as well? If we're treat them as something different, we could also choose to not catch them.

That's where it gets awkward. When an exception happens before runner even starts running the flow, what's the pair representing? I'd go with not catching them at all. And since it is happening outside the flow it's easy to debug

src/state_flow/core.clj
caioaao: wouldn't that make us lose info on the actual sub-flow where the exception happened?

dchelimsky: `description-stack` is no longer a thing on master. You can use [`current-description`](https://github.com/nubank/state-flow/blob/0d7354d29f5479abb61dfe9dba22326b10d7894b/src/state_flow/core.clj#L61) instead.

caioaao: @dchelimsky just pushed a commit using `current-description`, but it seems it was made to be used inside flows only and doesn't look "right" outside it?

caioaao: just pushed another change splitting the "pure" part of `current-description` from the monadic part so it can be used on the state directly (private `state->current-description` fn)

dchelimsky:
@caioaao I'd like to get this finailzed and merged this week. We've waited long enough that there are now conflicts. 3 options to proceed:

1. you resolve those and make any last minute adjustments after one last review (pending)
2. you give me write on your fork and I'll take care of that
3. I pull your commits from this branch onto a new branch in the nubank repo and take care of it there

What's your preference?

caioaao:
I can resolve the conflicts today, but I don't know what was holding it up
before the conflicts appeared. Let me know if there's anything else I need
to do after resolving the conflicts.

dchelimsky:
> but I don't know what was holding it up before the conflicts appeared.

Nothing specific. It's a significant change and there's a lot of other stuff going on. Now that you've got it up to date, I'll try to get the final review and hopefully merge in the next day or two.

dchelimsky:
Seems like there are still a couple of loose ends here:

* tests for exception within cleanup and within the runner
* deprecate `run!`

There is also a bunch of noise now when running tests that I'd taken some steps to get rid.

If everybody is on board, I'd say let's merge this as/is, and I can pick up the remaining adjustments before releasing.

/cc @caioaao @philomates @sovelten @nubank/eng-prod-test

dchelimsky:
I added some commits on a new branch: https://github.com/nubank/state-flow/commits/dc/post-62-adjustments. @caioaao if you like them as/is, you're welcome to merge them to your branch and then we can merge this PR. Otherwise I'll make a separate PR and wait to merge that before releasing.

caioaao:
I'd keep the behavior of `run!` unchanged so we don't break code that uses
custom runners based on that, but other than that lgtm.

dchelimsky:
>> I'd keep the behavior of `run!` unchanged so we don't break code that uses custom runners based on that, but other than that lgtm.

@caioaao I don't believe I changed the behavior of `run!`, just its implementation. I opened #101 with my additional changes. If you think there's a behavior change, please comment on that PR.

PR #63, opened 2020-01-16
make `return` constructable/def'able

dchelimsky:
I was writing up some intro docs and tried to evaluate `(state/return 37)` and got an exception. I think `return`, like the other primitives, should be constructable and def'able.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## [DEV]
+## [2.1.5]
+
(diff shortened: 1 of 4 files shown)


dchelimsky:
@philomates I bumped the project version and amended the Changelog. Would you please drop a release? Thx.

philomates:
@dchelimsky :+1:  released as `2.1.5`

dchelimsky:
Thanks @philomates

PR #67, opened 2020-01-21
there's a core fn for that

dchelimsky:


dchelimsky:
@philomates @sovelten this refactoring has zero outward impact, so just quietly merging. No need for changelog or release.

PR #69, opened 2020-01-21
Change `defflow` to also define the actual flow

caioaao:
The naming and what it does have a big mismatch IMO. In the current implementation `defflow` actually defines a test that runs the flow, and theflow is never actually defined.

This proposes also defining the actual flow in a variable named `name` and have the test name reflect what it actually is: running the flow defined by `name`.

But this is a breaking change. I couldn't think of a way of fixing this another way though.

### DIFF
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -41,5 +41,8 @@
   (let [[parameters & flows] (if (map? (first forms))
                                forms
                                (cons {} forms))]
-    `(ctest/deftest ~name
-       (core/run* ~parameters (core/flow ~(str name) ~@flows)))))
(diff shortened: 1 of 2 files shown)


philomates:
I agree that this is conflated and unfortunate! 
Do you have examples of when you would want to define a flow and also run it (in the clojure.test sense)?

caioaao:
Not off the top of my head. I had a use case for this before opening the PR that wasn't actually necessary, but I decided to open the PR anyway for semantics. But now that you mention it, this is in the `clojure.test` context, so it makes sense to just define a test (like `defspec` from `clojure.test.check`). I'm closing this PR :wave:

Issue #66, opened 2020-01-20
report failed assertions in return value of `run` function


dchelimsky:
When a `match?` assertion fails, the failure info is printed to the repl, and the return value of `run` doesn't include any information about it:

```clojure
(run
  (flow "interact with db"
    [db (state/gets :db)]
    (state/wrap-fn #(register-user db {:name "Phillip"}))
    (match? "user got added"
            (state/wrap-fn #(fetch-users db))
            #{{:name "Philip"}})) ;; <- different spelling
  {:db (atom {:users #{}})})
;; => #<Pair [#{{:name "Phillip"}} {:db #atom[{:users #{{:name "Phillip"}}} 0x4c2a27a6]}]>

;; --- printed to repl ---
;; FAIL in () (form-init13207122878088623810.clj:256)
;; interact with db (line 253) -> user got added
;; expected: (match? #{{:name "Philip"}} actual__12555__auto__)
;; actual: #{{:name (mismatch "Philip" "Phillip")}}
;; --- /printed to repl ---
```

This is fine if you're typing in the repl, but if you're using an dev environment in which you type in source files and send expressions to a connected repl, things get a bit disconnected.

I'm not sure about the format yet, but I think we should report assertion failures in the state, e.g.

```clojure
#<Pair [#{{:name "Phillip"}}
        {:db #atom[{:users #{{:name "Phillip"}}} 0x4c2a27a6]
         :failed-assertions
         [{:location "form-init13207122878088623810.clj:256"
            :name "interact with db (line 253) -> user got added"
            :expected (match? #{{:name "Philip"}} actual__12555__auto__)
            :actual #{{:name (mismatch "Philip" "Phillip")}}}]}]>
```

Then, perhaps, the `run` function could inspect `state` for failures and include them in the return - possibly wrapping them in a `Failure` object to align with runtime exceptions.

philomates:
This sounds like the idea that @sovelten has been pushing forward that test frameworks should avoid being effectul until the very last moment possible. As in, they should decouple evaluating tests with reporting tests. 

I think implementing this here might be a nice way to see whether it is a worthwhile idea to push forward in an actual test framework, either by adapting an existing one or writing a new one.

This should be relatively straightforward to implement with `matcher-combinators` because currently `state-flow` uses the standalone `match?` construct to do probing and only uses the clojure.test version for failure reporting.

[...]

PR #70, opened 2020-01-23
capture match results and add them to metadata of state

dchelimsky:
addresses #66 

What was done:
* match? now saves test report as metadata for each test done with match?
* instead of making a clojure.test assertion with match?, instead we make the assertions on `defflow` macro by calling `clojure.test/report` with the test report metadata that we gathered after the flow is run. This approach was used to preserve line and file information from when the assertion was made.

This way we are decoupling "making assertions" from reporting them.

To plugin other test frameworks and/or test runners, all that is needed is for the tool to gather and understand state-flow test report metadata.

Other "asserters" different from `match?` can be theoretically made as long as they push their report to metadata following our data structure (which is still internal/subject to change for the moment)

TODO:
* Create a spec for the report data structure

### DIFF
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -17,6 +17,7 @@
     `(~'state-flow.assertions.matcher-combinators/match? ~expected ~actual ~params*)))
 
 (defn clojure-test-report
+  "Internal use only, subject to change"
   [{:match/keys [result expected actual]
     :flow/keys [description-stack]
     :as assertion-report}]
@@ -25,8 +26,8 @@
      :message message
      :expected expected
      :actual actual
-     :file (-> assertion-report :flow/description-stack last :file)
-     :line (-> assertion-report :flow/description-stack last :line)}))
+     :file (-> description-stack last core/description->file)
+     :line (-> description-stack last :line)}))
 
 (defmacro defflow
   {:doc "Creates a flow and binds it a Var named by name"
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -1,7 +1,6 @@
 (ns state-flow.cljtest
   (:require [clojure.test :as t]
             [state-flow.assertions.matcher-combinators]
-            [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
             [state-flow.core :as core]
             [state-flow.probe :as probe]))
 
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -8,14 +8,20 @@
             [state-flow.state :as state]))
 
 (defn match-probe
-  "Returns the right value returned by probe/probe."
+  "Returns a map of :value (bound to the right value of the result of
+  probe/probe) and :match-results."
   ([state matcher]
    (match-probe state matcher {}))
   ([state matcher params]
-   (m/fmap second
-           (probe/probe state
-                        #(matcher-combinators/match? (matcher-combinators/match matcher %))
-                        params))))
+   (let [match-results (atom [])]
+     (m/fmap (fn [[_ r]] {:value r
+                          :match-results @match-results})
+             (probe/probe state
+                          (fn [v]
+                            (let [res (matcher-combinators/match matcher v)]
+                              (swap! match-results conj res)
+                              (matcher-combinators/match? res)))
+                          params)))))
 
 (defmacro match?
   "Builds a clojure.test assertion using matcher-combinators.
@@ -30,12 +36,17 @@
    {:description match-desc
     :caller-meta (meta &form)}
    `(m/do-let
-     [flow-desc# (core/current-description)
-      actual#    (if (state/state? ~actual)
-                   (match-probe ~actual ~expected ~params)
-                   (state/return ~actual))]
-     (state/wrap-fn #(t/testing flow-desc# (t/is (~'match? ~expected actual#))))
-     (state/return actual#))))
+     [flow-desc#    (core/current-description)
+      probe-result# (if (state/state? ~actual)
+                      (match-probe ~actual ~expected ~params)
+                      (match-probe (state/return ~actual)
+                                   ~expected
+                                   {:sleep-time 0 :times-to-try 1}))
+      :let [value#         (:value probe-result#)
+            match-results# (:match-results probe-result#)]]
+     (state/wrap-fn #(t/testing flow-desc# (t/is (~'match? ~expected value#))))
+     (core/modify-meta update :match-results (fnil conj []) match-results#)
+     (state/return value#))))
 
 (defmacro defflow
   {:arglists '([name & flows]
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -3,7 +3,8 @@
             [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
             [state-flow.assertions.matcher-combinators]
             [state-flow.core :as core]
-            [state-flow.probe :as probe]))
+            [state-flow.probe :as probe]
+            [matcher-combinators.printer :as matcher-combinators.printer]))
 
 (defmacro ^:deprecated match?
   "DEPRECATED. Use state-flow.assertions.matcher-combinators/match? instead. "
@@ -16,8 +17,16 @@
                        params)]
     `(~'state-flow.assertions.matcher-combinators/match? ~expected ~actual ~params*)))
 
-(defn clojure-test-report
-  "Internal use only, subject to change"
+(defn- tag-for-pretty-printing [actual-summary result]
+  (with-meta {:summary      actual-summary
+              :match-result result}
+    {:type ::mismatch}))
+
+(defmethod clojure.core/print-method ::mismatch [{:keys [match-result]} out]
+  (binding [*out* out]
+    (matcher-combinators.printer/pretty-print match-result)))
+
+(defn- clojure-test-report
   [{:match/keys [result expected actual]
     :flow/keys [description-stack]
     :as assertion-report}]
@@ -25,7 +34,11 @@
     {:type (case result :match :pass :mismatch :fail)
      :message message
      :expected expected
-     :actual actual
+     :actual (if (:mismatch/detail assertion-report)
+               (tag-for-pretty-printing
+                (list '~'not (list 'match? expected actual))
+                (:mismatch/detail assertion-report))
+               actual)
      :file (-> description-stack last core/description->file)
      :line (-> description-stack last :line)}))
 
@@ -41,5 +54,5 @@
        (let [[ret# state#] (core/run* ~parameters (core/flow ~(str name) ~@flows))
              assertions# (get-in (meta state#) [:test-report :assertions])]
          (doseq [assertion-data# assertions#]
-           (t/report (clojure-test-report assertion-data#)))
+           (t/report (#'clojure-test-report assertion-data#)))
          [ret# state#]))))
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -39,7 +39,7 @@
                                (cons {} forms))]
     `(t/deftest ~name
        (let [[ret# state#] (core/run* ~parameters (core/flow ~(str name) ~@flows))
-             test-report# (get (meta state#) :test-report)]
-         (doseq [assertion-data# (:assertions test-report#)]
+             assertions# (get-in (meta state#) [:test-report :assertions])]
+         (doseq [assertion-data# assertions#]
            (t/report (clojure-test-report assertion-data#)))
          [ret# state#]))))
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -1,10 +1,10 @@
 (ns state-flow.cljtest
   (:require [clojure.test :as t]
-            [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
+            [matcher-combinators.printer :as matcher-combinators.printer] ;; to register clojure.test assert-expr for `match?`
+            [matcher-combinators.test]
             [state-flow.assertions.matcher-combinators]
             [state-flow.core :as core]
-            [state-flow.probe :as probe]
-            [matcher-combinators.printer :as matcher-combinators.printer]))
+            [state-flow.probe :as probe]))
 
 (defmacro ^:deprecated match?
   "DEPRECATED. Use state-flow.assertions.matcher-combinators/match? instead. "
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -1,5 +1,6 @@
 (ns state-flow.cljtest
   (:require [clojure.test :as t]
+            [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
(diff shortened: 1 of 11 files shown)


src/state_flow/cljtest.clj
```diff
@@ -8,31 +8,42 @@
...
   (let [form-meta (meta &form)]
     `(core/flow ~match-desc
-       [flow-desc# (core/current-description)
-        actual#    (if (state/state? ~actual)
-                     (match-probe ~actual ~checker ~params)
-                     (state/return ~actual))]
-       (state/wrap-fn #(do (match-expr flow-desc# actual# ~checker ~form-meta)
-                           actual#)))))
+       [flow-desc#    (core/current-description)
+        probe-result# (if (state/state? ~actual)
```
dchelimsky: Leaning on `match-probe` even when we have an explicit value so that we always get the match result added to metadata.

philomates:
Very cool!
I wonder if it will be annoying to show the matcher-combinator results in a pretty-printed manner when interacting with results via the repl as mentioned in https://github.com/nubank/state-flow/issues/66

dchelimsky:
> I wonder if it will be annoying to show the matcher-combinator results in a pretty-printed manner when interacting with results via the repl as mentioned in #66

Annoying how?

philomates:
> > I wonder if it will be annoying to show the matcher-combinator results in a pretty-printed manner when interacting with results via the repl as mentioned in #66
> 
> Annoying how?

Perhaps a poor choice of word :) I guess I meant "how much work", like can we make sure that the `printer` ns is referenced (https://github.com/nubank/matcher-combinators/blob/master/src/cljc/matcher_combinators/printer.cljc) or is something more needed?

dchelimsky:
@philomates this is what it looks like with this PR:

```clojure
user> (map meta (state-flow/run (cljtest/match? "by any other name ..." "soccer" #"futebol")))

FAIL in () (form-init17791069151438309800.clj:167)
by any other name ... (line 167)
expected: (match? #"futebol" (:value probe-result__12593__auto__))
  actual: (mismatch #"futebol" "soccer")

(nil
 {:top-level-description "by any other name ...",
  :description-stack [],
  :match-results
  [[{:matcher-combinators.result/type :mismatch,
     :matcher-combinators.result/value
     {:expected #"futebol", :actual "soccer"},
     :matcher-combinators.result/weight 1}]]})
```

Are you thinking of something other than `{:expected #"futebol" ...}`

dchelimsky:
> (:value probe-result__12593__auto__) has to go (replaced by "soccer" in this case)

Actually, the mismatch comment below shows the value "soccer" - this is a clojure.test thing and is fine as/is IMO.

dchelimsky:
Let's hold on this, given [the conversation about decoupling matcher-combinators from state-flow](https://github.com/nubank/state-flow/issues/66).

src/state_flow/cljtest.clj
```diff
@@ -1,38 +1,52 @@
...
+  "Returns a map of :value (bound to the right value of the result of
+  probe/probe) and :match-results."
   ([state matcher]
    (match-probe state matcher {}))
   ([state matcher params]
-   (m/fmap second
-           (probe/probe state
-                        #(matcher-combinators/match? (matcher-combinators/match matcher %))
-                        params))))
+   (let [match-results (atom [])]
```
sovelten: why the atom?

dchelimsky: It's stateful. Each try in the probe appends a map to the vector in the atom. Make sense?

sovelten: I think so, but does it need to be stateful?
How about we change the behavior of probe to return all the results instead?
I think this is actually pretty clever and the function is still referentially transparent, but I have a natural aversion to statefulness so I was wondering if there is a better way.

philomates: I had the same reaction when I saw the atom but couldn't think of a pure implementation that wouldn't be overly complicated in comparison

dchelimsky: We could handle this in probe, but I don't see how it's not stateful. We're starting with an empty list and adding things to it as we iterate.

sovelten: Idea for modification of retry:
```clojure
(defn ^:private retry
  "Tries at most n times, returns a vector with true and first element that succeeded
  or false and result of the first try"
  [times pred? lazy-seq]
  (let [[tried remaining] (split-with (complement pred?) (take times lazy-seq))]
    (if (empty? remaining)
      [false tried]
      [true  tried])))
```

dchelimsky: @sovelten `split-with` will try (and pause) `times` even if the first try succeeds, slowing down test runs. Also, this would be a breaking change of `probe`.

We could move the implementation from `match-probe` to `probe` and store the results on the context state there. WDYT of that?

dchelimsky: Also, the value of `tried` in your example is the return value of `(state/get)` on the first arg to `probe`, not the match result map:

```clojure
(state-flow/run (probe/probe test-helpers/get-value-state #(= 2 %)
                               {:sleep-time   250
                                :times-to-try 5})
    {:value (atom 0)})
;; => #<Pair [[false 0 (nil 0 0 0 0 0)] {:value #atom[0 0xc3915c9]}]>
```

dchelimsky:
I extracted out the refactoring in this PR to #76, because I'd like to see that merged regardless of what we decide about behaviour.

src/state_flow/cljtest.clj
sovelten: I was hoping the implementation of `split-with` would be smart enough not to evaluate the rest of the seq, but it literally uses take-while and drop-while, it doesn't actually run `times`, but it runs twice the amount... oh well.

I'm gonna think of something. but maybe we can stick with that solution meanwhile.

I'm ok with breaking `probe`, I don't think it is actually used directly. But it is always possible to deprecate and think of a better design.

sovelten: With https://github.com/nubank/state-flow/pull/78 you could use `probe*` instead and eliminate the atom

src/state_flow/core.clj
```diff
@@ -259,22 +265,23 @@
...
            cleanup                identity
            runner                 run
            fail-fast?             false
            before-flow-hook       identity
            on-error               (comp throw-error!
                                         log-error
-                                        (filter-stack-trace default-stack-trace-exclusions))}
-    :as   opts}
+                                        (filter-stack-trace default-stack-trace-exclusions))
+           assert-with-clojure-test? false}}
```
dchelimsky, line 277: 2 concerns:

* this makes clojure.test special
* as support for new frameworks emerges, we'll need a new config option per framework.
* I _think_ it's a breaking change, since users, at least a Nubank, all have some sort of custom `defflow` that would have to change.

Alternate proposal:

```clojure
assert-with 'clojure.test
```

This would leave the default intact and open us up for new frameworks. WDTY?

Then we could probably do away with the `assert-with-clojure-test?` function as well, and just check something like this in `state-flow.assertions.matcher-combinators`:

```clojure
(state/gets (comp :assert-with meta))
```

sovelten: Clojure test is special. It is special because it is clunky. I do not expect to add this option for any other test framework. Maybe I'm being too optimistic?

It is not a breaking change as long as people use `defflow` for clojure.test which they should. Even at Nubank the custom defflows are based on this one.

I guess we could change to `:assert-with 'clojure.test`, but I just don't want to pass the message that this option should be extended. It shouldn't. Other test frameworks should use the test report metada after the flow ran.

dchelimsky: What you're saying makes me wish we could do this _all_ in the context of the `state-flow.cljtest` namespace, and leave everything else ignorant of it. Just brainstorming here, but maybe the `state-flow.assertions.matcher-combinators` namespace could look for an `assertion-flow` (or similar) in the state (or on meta) and the `state-flow.cljtest` ns could provide it in its `defflow`. WDYT of that direction?

sovelten: Yeah, I think what you are suggesting is actually some sort of test-assertion-hook, a function in the form of `(fn [test-data] make-assertion)` that we can pass as a custom parameter.

I'm afraid we would have the same problem of incorrect line-numbering though. But if you think it could work, we can try. The problem of getting the correct line numbering on clojure.test is actually very confusing to me.

sovelten: The `is` thingy need to be part of the actual code when the macro is expanded, it cannot be called indirectly.

sovelten: From what I understand clojure.test is going to get the line number from the current thread stacktrace, it's very weird:
https://github.com/clojure/clojure/blob/28efe345d5e995dc152a0286fb0be81443a0d9ac/src/clj/clojure/test.clj#L348

Even considering that we know the correct line numbers we cannot build an assertion with them at runtime (unless maybe we bind the `.getFileName` and `.getLineNumber` to return what we want? Sounds very hacky but might be doable)

sovelten: It works by changing the call of `stacktrace-file-and-line` to return the file name and line number we want. But that is a with-redefs, so not thread safe, not cool. I'm out of ideas.

dchelimsky: I have an idea that I'm trying to work out. Can you give me a day?

dchelimsky: My idea won't work. That said, I think we could (should) still keep this constrained to the `cljtest` and `matcher-combinators` namespaces. No need to gum up the API for this.

Here's a proposal: https://github.com/nubank/state-flow/commit/7b72cf3ee4b30cde6d6c928ed0f88726a4b5f183

sovelten: I agree with the proposal, but I would like to try yet another thing. My idea is to call `report` directly:
https://github.com/clojure/clojure/blob/28efe345d5e995dc152a0286fb0be81443a0d9ac/src/clj/clojure/test.clj#L375

dchelimsky: I love that idea because that lets us pass around an assertion function, but I think the cost would be crafting the message, including file/line number info in `match?`, right? Seems like a good trade-off. Also means that message stands a better chance of being consistent across testing frameworks.

sovelten: We already have file and line number on our test report, so no big deal. The trade-off that I see is that we are relying on clojure.test internal data structure, I don't see it changing anytime soon though :thinking:

dchelimsky: That could be resolved with an adapter if/when the time comes. It's all internal.

dchelimsky: Unless you plan to use it, I'm going to delete the branch w/ my proposal (just to clean up).

sovelten: It needs some polishing, but this is the version that relies on report https://github.com/nubank/state-flow/pull/70/commits/f5812eaa34990484e41a08923dd1d3960bcc40a7

dchelimsky: Looking cleaner! I'd still like to get rid of `assert-with-clojure-test?` - is that in your plan?

sovelten: Yes, I also want to make some tests by binding the output of the clojure.test just to make sure reporting is still the same
I'll ask for another review when it is done.

I also would like to hear @philomates take on this

dchelimsky: > Yes, I also want to make some tests by binding the output of the clojure.test just to make sure reporting is still the same

I _think_ https://github.com/nubank/state-flow/blob/0164452705f4abb66cc8f08c0245ea4ebf9a0768/test/state_flow/test_helpers.clj#L36 already does what you want

src/state_flow/cljtest.clj
```diff
@@ -24,4 +37,8 @@
                                forms
                                (cons {} forms))]
     `(t/deftest ~name
-       (core/run* ~parameters (core/flow ~(str name) ~@flows)))))
+       (let [[ret# state#] (core/run* ~parameters (core/flow ~(str name) ~@flows))
+             test-report# (get (meta state#) :test-report)]
+         (doseq [assertion-data# (:assertions test-report#)]
```
philomates, line 42: perhaps you could right from the start do `(get-in (meta state#) [:test-report :assertions])`

test/state_flow/assertions/matcher_combinators_test.clj
```diff
@@ -94,10 +99,11 @@
...
         (is (match? {:match/result :mismatch} flow-ret)))
-      (testing "reports match-results to clojure.test"
-        (is (match? {:matcher-combinators.result/type  :mismatch
-                     :matcher-combinators.result/value {:expected 2 :actual 0}}
-                    (-> report-data :actual :match-result))))))
+      (testing "pushes test report to metadata"
+        (is (match? {:match/result  :mismatch
+                     :match/expected 2
+                     :match/actual 0}
+                    (-> (meta flow-state) :test-report :assertions first))))))
```
philomates, line 106: can you use a `get-in` like above (or use `->` above at https://github.com/nubank/state-flow/pull/70/files#diff-ee191a9ed7e370b3ecb731f245cd5a2ca14de431ecfe680ca4d5a18de55695ebR88)

src/state_flow/core.clj
```diff
@@ -55,13 +55,13 @@
...
-  [strs]
-  (->> strs
+(defn format-description
+  [stack]
+  (->> stack
        (map format-single-description)
        (str/join " -> ")))
 
-(defn ^:private description-stack [s]
+(defn description-stack [s]
```
philomates, line 64: do we want docstrings here that say they are internal functions?

philomates (approved):
:clap: super cool!

src/state_flow/core.clj
sovelten: just to be safe, yes
but there are cases where an user may want to have the full description data (unformatted), maybe this should be part of the api, but not the scope of this PR

src/state_flow/cljtest.clj
```diff
@@ -15,6 +16,19 @@
                        params)]
     `(~'state-flow.assertions.matcher-combinators/match? ~expected ~actual ~params*)))
 
+(defn clojure-test-report
```
dchelimsky, line 19: How about making this private and referencing it as `#'clojure-test-report` in the macro below?

src/state_flow/cljtest.clj
```diff
@@ -15,6 +16,19 @@
...
+  "Internal use only, subject to change"
+  [{:match/keys [result expected actual]
+    :flow/keys [description-stack]
+    :as assertion-report}]
+  (let [message (core/format-description description-stack)]
+    {:type (case result :match :pass :mismatch :fail)
+     :message message
+     :expected expected
+     :actual actual
```
philomates, line 28: If you want the colorized, in-line mismatch output of `matcher-combinators` here you need to register the pretty-printer over the `actual`

So roughly:
```clojure
     :actual   (if (:mismatch/detail assertion-report)
                 (tag-for-pretty-printing
                   (list '~'not (list 'match? expected actual))
                   (:mismatch/detail assertion-report))
                 actual)
```
where you have the following above
```clojure
(defn tag-for-pretty-printing [actual-summary result]
  (with-meta {:summary      actual-summary
              :match-result result}
    {:type ::mismatch}))

(defmethod clojure.core/print-method ::mismatch [{:keys [match-result]} out]
  (binding [*out* out]
    (printer/pretty-print match-result)))
```

philomates (approved):

PR #71, opened 2020-01-24
match? -> expect

dchelimsky:
This PR represents a proposal for how to deal with the argument ordering of `match?`: deprecate it and introduce a new `expect` function e.g.

```clojure
(expect <expected> <actual> <optional-params>)
```

In addition to introducing the function, this PR also includes a rudimentary rewrite script which was used to convert `match?` to `expect` in the `cljtest_test` namespace.

NOTE: this is initially based off the `dc/add-match-results-to-state-meta` branch, which is the PR branch for #70. That should be merged first.

### DIFF
--- a/dev/state_flow/rewrite/match_to_expect.clj
+++ b/dev/state_flow/rewrite/match_to_expect.clj
@@ -1,0 +1,71 @@
+(ns state-flow.rewrite.match-to-expect
+  (:require [clojure.java.io :as io]
+            [rewrite-clj.zip :as z]))
+
+(def cljtest-test-src
+  (delay (z/of-file "test/state_flow/cljtest_test.clj")))
+
+(def data "(state-flow/run (cljtest/match? \"DESC\" 'actual 'expected) {:initial :state})")
+
+(def ^:dynamic *match?-symbol* 'match?)
+(def ^:dynamic *expect-symbol* 'expect)
+
+(defn match?->expect
+  "Given a match? expression e.g.
+
+    (match? <description> <actual> <expected>)
+    ;; or
+    (match? <description> <actual> <expected> <params>)
+
+  Returns an expect expression e.g.
+
+    (expect <expected> <actual>)
+    ;; or
+    (expect <expected> <actual> <params>)
+
+  If there is an exception, printlns the expression so you can
+  find and handle it manually."
+  [expect-sym match-exp]
+  (try
+    (let [match-sym (-> match-exp z/down)
+          actual (-> match-sym
+                     z/right
+                     z/right)
+          expected (-> actual z/right)]
+      (-> match-sym
+          (z/replace expect-sym)
+          z/right
+          z/remove
+          z/right
+          (z/replace (z/node expected))
+          z/right
+          (z/replace (z/node actual))))
+    (catch Exception e
+      (println "Error processing " (z/string match-exp))
+      match-exp)))
+
+(defn match?-expr? [match?-sym node]
+  (when (= :list (z/tag node))
+    (when-let [op (z/down node)]
+      (= match?-sym (z/value op)))))
+
+(defn rewrite-match?->expect [match?-sym expect-sym path]
+  (let [data (z/of-file path)]
+    (spit (io/file path)
+          (z/root-string
+           (loop [data data]
+             (let [updated (try
+                             (z/prewalk data
+                                        (partial match?-expr? match?-sym)
+                                        (partial match?->expect expect-sym))
+                             (catch Exception _ data))]
+               (if (z/rightmost? updated)
+                 updated
+                 (recur (z/right updated)))))))))
+
+(comment
+  (rewrite-match?->expect 'cljtest/match?
+                          'cljtest/expect
+                          "test/state_flow/cljtest_test.clj")
+
+  )
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -30,22 +30,42 @@
                               (matcher-combinators/match? res)))
                           params)))))
 
+(defn expect*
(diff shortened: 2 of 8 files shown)


src/state_flow/cljtest.clj
```diff
@@ -30,22 +30,42 @@
...
+                                    (:value probe-result#))))))
+
+(defmacro expect
+  "Builds a clojure.test assertion using matcher combinators.
+
+  - expected can be a matcher-combinators matcher or a literal value
+    - literals will be used to infer default matchers
+  - actual can be a state monad or a literal value"
+  [expected actual & [params]]
+  (expect* {:description "expect"
```
dchelimsky: It would be great to update matcher-combinators to print themselves in friendly ways so the descriptions could read `expect 37` for the literal 37 and `expect re-find #"some-regex"` for `(matchers/regex #"some-regex")`

src/state_flow/cljtest.clj
```diff
@@ -30,22 +30,42 @@
...
+                (state/wrap-fn #(do (match-expr flow-desc# (:value probe-result#) ~expected ~form-meta)
+                                    (:value probe-result#))))))
+
+(defmacro expect
+  "Builds a clojure.test assertion using matcher combinators.
+
+  - expected can be a matcher-combinators matcher or a literal value
+    - literals will be used to infer default matchers
+  - actual can be a state monad or a literal value"
+  [expected actual & [params]]
```
philomates: there is no way to get an optional desc as well, right?
I guess adding a `:description` entry in the `params` map works well enough here

src/state_flow/cljtest.clj
```diff
@@ -30,22 +30,42 @@
...
+                (core/modify-meta update :match-results (fnil conj []) (:match-results (meta probe-result#)))
+                (state/wrap-fn #(do (match-expr flow-desc# (:value probe-result#) ~expected ~form-meta)
+                                    (:value probe-result#))))))
+
+(defmacro expect
+  "Builds a clojure.test assertion using matcher combinators.
+
+  - expected can be a matcher-combinators matcher or a literal value
+    - literals will be used to infer default matchers
+  - actual can be a state monad or a literal value"
```
philomates: maybe `state monad` -> `flow (state monad)` or `state monad flow`?

src/state_flow/cljtest.clj
```diff
@@ -30,22 +30,42 @@
...
+                                 (match-probe ~actual ~expected ~params)
+                                 (match-probe (state/return ~actual)
+                                              ~expected
+                                              {:sleep-time 0 :times-to-try 1}))]
+                (core/modify-meta update :match-results (fnil conj []) (:match-results (meta probe-result#)))
+                (state/wrap-fn #(do (match-expr flow-desc# (:value probe-result#) ~expected ~form-meta)
+                                    (:value probe-result#))))))
+
+(defmacro expect
+  "Builds a clojure.test assertion using matcher combinators.
```
philomates: `matcher combinators` -> `matcher-combinators`

dev/state_flow/rewrite/match_to_expect.clj
```diff
@@ -0,0 +1,71 @@
...
+                     z/right)
+          expected (-> actual z/right)]
+      (-> match-sym
+          (z/replace expect-sym)
+          z/right
+          z/remove
+          z/right
+          (z/replace (z/node expected))
+          z/right
+          (z/replace (z/node actual))))
```
philomates: ah zippers :)

src/state_flow/cljtest.clj
dchelimsky: How about:

```
- actual can be a literal value, a primitive, or a flow
;; or
- actual can be a literal value, a primitive step, or a flow
```

We've gotten into the habit of calling primitives flows, and I find that confusing. I'd like to keep that distinction visible and consistent. Maybe we can add something to the readme about the Composite pattern, and the fact that the runner sees primitives and flows as the same thing, etc.

dev/state_flow/rewrite/match_to_expect.clj
dchelimsky: Yeah - this was not fun, but quite satisfying in a caveman sort of way when I got it to work :)

src/state_flow/cljtest.clj
dchelimsky: One of the goals here is to obviate the need for a description most of the time. If "expect 37" isn't clear enough, you can always wrap it in a flow, which does require a description. WDYT?

dev/state_flow/rewrite/match_to_expect.clj
sovelten: Nice! This is the rewrite-clj library I've been talking about.

This was the script I did for converting from `verify` to `match?`:
https://gist.github.com/sovelten/67583e8e3b66e584fee12caf62b8c191

Also deeply satisfying after I got it working

dchelimsky: Check out [the beginnings of a tool to convert facts to cljtest](a https://github.com/nubank/clojure-refactoring/blob/dc/midje-to-cljtest/midje-to-cljtest/src/nubank/rewrite/fact_to_testing.clj), too.

There are a couple of general purpose functions in there (and more I'm working on, like `drop-z` and `take-z` - guess what they do!?!) that I think we can extract to an extension lib for rewrite-clj (unless they accept PRs, but there have been no commits on that lib for about 15 months).

sovelten:
As we discussed the other day, we could have `expect` and some other function `expect-that` that accepts a string description.

But we might want to hold this change till after we think about a clearer plan forward.

dchelimsky:
We're going to create a new `match?` function in a new namespace instead. Closing.

PR #73, opened 2020-01-28
More helper functions on state ns

caioaao:
To avoid having ppl complaining about using a monad :speak_no_evil:

### DIFF
--- a/src/state_flow/state.clj
+++ b/src/state_flow/state.clj
@@ -1,5 +1,5 @@
 (ns state-flow.state
-  (:refer-clojure :exclude [eval get])
+  (:refer-clojure :exclude [eval get -> ->>])
   (:require [cats.context :as ctx :refer [*context*]]
             [cats.core :as m]
             [cats.data :as d]
@@ -59,27 +59,27 @@
 (util/make-printable (type error-context))
 
 (defn get
-  "Returns the equivalent of (fn [state] [state, state])"
+  "Evaluates the equivalent of (fn [state] [state, state])"
   []
   (state/get error-context))
 
 (defn gets
+  "Evaluates to the equivalent of (fn [state] [state, (f state)])"
   [f]
-  "Returns the equivalent of (fn [state] [state, (f state)])"
   (state/gets f))
 
 (defn put
-  "Returns the equivalent of (fn [state] [state, new-state])"
+  "Evaluates to the equivalent of (fn [state] [state, new-state])"
   [new-state]
   (state/put new-state error-context))
 
 (defn modify
-  "Returns the equivalent of (fn [state] [state, (swap! state f)])"
+  "Evaluates to the equivalent of (fn [state] [state, (swap! state f)])"
   [f]
   (state/swap f error-context))
 
 (defn return
-  "Returns the equivalent of (fn [state] [v, state])"
+  "Evaluates to the equivalent of (fn [state] [v, state])"
   [v]
   (m/return error-context v))
 
@@ -95,13 +95,18 @@
                  (d/pair (my-fn) s))
                error-context))
 
-(defn bind
-  "Equivalent to `(flow \"\" [a mv] (return (f a)))`"
-  [mv f]
-  (m/bind mv f))
+(defmacro ->
+  "Analog to Clojure's `->`, but operating on expressions that result in a state instance."
+  [mv & exprs]
+  `(m/->= ~mv ~@exprs))
+
+(defmacro ->>
+  "Analog to Clojure's `->>`, but operating on expressions that result in a state instance."
+  [mv & exprs]
+  `(m/->>= ~mv ~@exprs))
 
 (defn fmap
-  "Equivalent to `(flow \"\" [a mv] (f a))`"
+  "Evaluates to the equivalent of (mlet [a mv] (state/return (f a)))"
   [f mv]
   (m/fmap f mv))
 
--- a/src/state_flow/state.clj
+++ b/src/state_flow/state.clj
@@ -95,6 +95,16 @@
                  (d/pair (my-fn) s))
                error-context))
 
+(defn bind
+  "Equivalent to `(flow \"\" [a mv] (return (f a)))`"
+  [mv f]
+  (m/bind mv f))
+
+(defn fmap
+  "Equivalent to `(flow \"\" [a mv] (f a))`"
+  [f mv]
+  (m/fmap f mv))
+
 (def state? state/state?)
 (def run state/run)
 (def eval state/eval)
--- a/src/state_flow/state.clj
+++ b/src/state_flow/state.clj
@@ -60,27 +60,27 @@
 (util/make-printable (type error-context))
 
 (defn get
-  "Evaluates the equivalent of (fn [state] [state, state])"
+  "Returns the equivalent of (fn [state] [state, state])"
   []
   (state/get error-context))
 
 (defn gets
-  "Evaluates to the equivalent of (fn [state] [state, (f state)])"
+  "Returns the equivalent of (fn [state] [state, (f state)])"
   [f]
   (state/gets f))
 
 (defn put
-  "Evaluates to the equivalent of (fn [state] [state, new-state])"
+  "Returns the equivalent of (fn [state] [state, new-state])"
   [new-state]
   (state/put new-state error-context))
 
 (defn modify
-  "Evaluates to the equivalent of (fn [state] [state, (swap! state f)])"
+  "Returns the equivalent of (fn [state] [state, (swap! state f)])"
   [f]
   (state/swap f error-context))
 
 (defn return
-  "Evaluates to the equivalent of (fn [state] [v, state])"
+  "Returns the equivalent of (fn [state] [v, state])"
   [v]
   (m/return error-context v))
 
@@ -103,10 +103,10 @@
   Usage:
 
   ```clojure
-  (defn inc [x] (state/return (inc x)))
+  (defn m-inc [x] (state/return (inc x)))
 
   (state/-> (state/return 1)
-            inc)
+            m-inc)
   ```"
   [mv & exprs]
   `(m/->= ~mv ~@exprs))
@@ -118,10 +118,10 @@
   Usage:
 
   ```clojure
-  (defn map2 [f vs] (state/return (map f vs)))
+  (defn m-map [f vs] (state/return (map f vs)))
 
   (state/->> (state/return [1 2 3])
-             (map2 inc))
+             (m-map inc))
   ```"
   [mv & exprs]
   `(m/->>= ~mv ~@exprs))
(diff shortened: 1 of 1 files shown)


src/state_flow/state.clj
```diff
@@ -95,6 +95,16 @@
...
 
+(defn bind
+  "Equivalent to `(flow \"\" [a mv] (return (f a)))`"
+  [mv f]
+  (m/bind mv f))
+
+(defn fmap
+  "Equivalent to `(flow \"\" [a mv] (f a))`"
+  [f mv]
+  (m/fmap f mv))
```
philomates: I find the names `fmap` and `bind` a little abstract, but I guess that is the point :) 
And it looks like they match the haskell names, so it is in line with https://github.com/nubank/state-flow/pull/42

philomates: these wrappers aren't really specific to the state-monad, but I guess this is the best place to put them

src/state_flow/state.clj
```diff
@@ -95,6 +95,16 @@
                  (d/pair (my-fn) s))
                error-context))
 
+(defn bind
+  "Equivalent to `(flow \"\" [a mv] (return (f a)))`"
+  [mv f]
+  (m/bind mv f))
+
+(defn fmap
+  "Equivalent to `(flow \"\" [a mv] (f a))`"
```
philomates: I think a prose description would also be helpful (along with tests that serve as examples)

src/state_flow/state.clj
caioaao: about the names: yeah, I couldn't find better names either, and at least people that used spec/core.check generators will be familiar with them.

about being specific to the state monad: they're not, but that's the point: passing the impression of an specific/unified API for dealing with state-flow's states and more advanced users can drop down to the monad/functors/etc api if they want to.

src/state_flow/state.clj
caioaao: I could use some help on this tbh :sweat_smile: . The equivalent is not even strictly true, but it was the best way I found to describe it with only state-flow's lingo

philomates: maybe adapting what is in cats?
https://github.com/funcool/cats/blob/master/src/cats/core.cljc#L82-L91

src/state_flow/state.clj
```diff
@@ -95,6 +95,16 @@
                  (d/pair (my-fn) s))
                error-context))
 
+(defn bind
+  "Equivalent to `(flow \"\" [a mv] (return (f a)))`"
```
dchelimsky: There's no argument `a` below, so I'm not clear what this means.

dchelimsky: Oh, never mind - we're binding the result of `mv` to `a`. This could definitely be made more clear for the monadic-impaired.

caioaao: Oh yeah. You not getting it on the first read is already convincing that this explanation is bad :)

dchelimsky: Can you show an example of how to use it?

caioaao: Sure! The biggest usage in my case is this: I have a function that creates an user and returns it:
```clojure
(defn user-signup! "Signs up using parameters and return user" [parameters]
  (flow "Creating complete user"
        [user (http/create! parameters)]
        (confirm-email! user)
        (cats/return user)))
```

Sometimes I need to have the returned user during a flow, like this
```clojure
(defflow bla
  ...
  [user (user-after-signup! {...})]
  ;; use unwrapped user value here
  )
```

But sometimes I just want to pass it to another function. Without the `bind` I'd have to do this:
```clojure
(defflow ble
  [user (user-after-signup! {...})
   foo (use-user! user)]
   ...
   )
```

But with bind:
```clojure
(defflow ble-2
  [foo (-> (user-after-signup! {...})
           (bind use-user!))]
   ...
  )
```

So basically a helper to compose functions that evaluate to monadic values (there's even an arrow macro equivalent on cats - `cats/->=` - but that's too much cognitive load for a testing library in my opinion)

dchelimsky: Thanks, that helps.

dchelimsky: OK, so the description is somewhere between "here's what this does/returns" and "here's how you use it", but not quite either. I'll follow up with suggestions.

Side note: the `user-signup!` function, I really want to discourage using `cats` directly. There is a `state-flow.state/return` function that calls into `cats.core/return` and I think that's what we should be using. WDYT?

dchelimsky: I think `bind` is a confusing name here because we're really applying a function to an unwrapped value, which we _then_ bind to the symbol on the left. You say elsewhere that `cats/->=` is too much cognitive overload, but I think this would be easier to grok than your example above:


```clojure
(defflow ble-2
  [foo (->= (user-after-signup! {...})
             use-user!))]
   ...
  )
```

Alternatively, we could name it something that relates more to state-flow, but still has the arrow in it, like `state/->`.

caioaao: Oh, now that you said it, I like the `state/->` and it'd make `bind` redundant. Looks a lot clearer and even more useful than just `bind` itself :+1:

sovelten:
I'm not really sure this should live on the `state` namespace or some other namespace.
For instance, you mention flows in the docs, but `state.clj` namespace is not aware of flow, it is the opposite: flow requires state.

How about moving to some other namespace like helpers.clj or something like that?

caioaao:
> I'm not really sure this should live on the `state` namespace or some other namespace.
> For instance, you mention flows in the docs, but `state.clj` namespace is not aware of flow, it is the opposite: flow requires state.
> 
> How about moving to some other namespace like helpers.clj or something like that?

I think that's more a reflection on the bad docstrings than actually being at the wrong place. I used flows here because I thought it was the more familiar concept I could use. We can revisit the placement after docstrings are better, wdyt?

src/state_flow/state.clj
```diff
@@ -59,27 +60,27 @@
 (util/make-printable (type error-context))
 
 (defn get
-  "Returns the equivalent of (fn [state] [state, state])"
+  "Evaluates the equivalent of (fn [state] [state, state])"
```
dchelimsky: Returns is correct. Why did you change this?

dchelimsky: What it returns is a state monad, which wraps a function like `(fn [state] [state, state])`, but there is no evaluation when you invoke `(get)`.

src/state_flow/state.clj
```diff
@@ -95,6 +96,47 @@
...
                error-context))
 
+(defmacro ->
+  "Analog to Clojure's `->`, but operating on expressions that result in a state
+  instance. An example:
+
+  Usage:
+
+  ```clojure
+  (defn inc [x] (state/return (inc x)))
```
dchelimsky: Please call this something other than a core clojure function :) Maybe `inc-state` or some such.

src/state_flow/state.clj
```diff
@@ -95,6 +96,47 @@
                  (d/pair (my-fn) s))
                error-context))
 
+(defmacro ->
+  "Analog to Clojure's `->`, but operating on expressions that result in a state
+  instance. An example:
+
+  Usage:
+
+  ```clojure
```
dchelimsky: I haven't seen this markdown syntax used within docstrings. What environment formats this for you? Not github, and not a REPL. Where else?

src/state_flow/state.clj
```diff
@@ -95,6 +96,47 @@
                  (d/pair (my-fn) s))
                error-context))
 
+(defmacro ->
```
sovelten: This is equivalent to `(fmap inc (state/return 1))`
I never use this kind of macro and I would not want to encourage anyone to use it. It is better to have one clear way to do it than several different ways.

If you want to apply several functions, `comp` works fine:
`(m/fmap (comp inc inc) (state/return 1))`

And there is no need to differentiate between `->` and `->>` as `comp` is agnostic to that and it receives functions as arguments and not pieces of code with an argument missing. No need to introduce thread macros to the monad world as thread macros are a bad design choice anyway.

dchelimsky: > thread macros are a bad design choice anyway

That's a pretty objective sounding statement for a subjective statement ;) Care to elaborate?

dchelimsky: I do agree, btw, with excluding `->` and `->>` here. Just want to understand why you think they're universally bad.

sovelten: I have just the thing: https://gist.github.com/sovelten/486602d2711ed251ffee41215b7e8dbc

caioaao: heh, that was a test. I knew you wouldn't like it @sovelten :joy: But `comp` doesn't work when each functions is already returning a monadic value. we'd need something like haskell's `>=` (which is ugly and confusing af) or re-add `bind` here (which is not the preferred way even in haskell). This, IMO, is a clear place where a new syntax can/should be introduced and what I did was just introducing one that would already be familiar to Clojure users

src/state_flow/state.clj
caioaao: I think cljdoc.org supports this, but now that you mention I don't know if it's true. Anyway, I'd say this is not the optimal way of adding examples (it's not testable as in Python), but I couldn't think of an alternative that would have it close to the function definition.

src/state_flow/state.clj
caioaao: Oops. I wasn't supposed to commit that, sorry :sweat: . I was rewording it from something I've read a long time ago about how the lingo should be different for REPL based languages, but I couldn't find it and wasn't sure about the discussion anymore.

src/state_flow/state.clj
dchelimsky: Just leave off the markdown (i.e. `\``clojure ... ```) and it's fine to have the Usage example here IMO.

src/state_flow/state.clj
sovelten: Ok, I get the idea now. But I don't see the benefit of yet another syntactic sugar option, it is just another thing for people to learn and get confused about. `do-let` already provides all the necessary flexibility and it mirrors Haskell do notation.

caioaao: You're right the library already provides enough flexibility, but it does lead to a pretty verbose code IMO. This might be a matter of preference/style, though.

src/state_flow/state.clj
caioaao: Just double-checked and cljdoc.org renders the markdown in docstrings. Might be better to leave as is then, no?

src/state_flow/state.clj
dchelimsky: @caioaao we want to be cautious about expanding the surface of state-flow, so we're not going to add these right now. We think you're not alone in wanting to reduce verbosity, and that an open source extension library would be appreciated by other users of state-flow, so that's definitely an option for you. Alternatively, you could just use the cats macros and functions directly (`flow` is just a wrapper for `do-let`). I'll add something to the README saying "For advanced users,`flow` is a wrapper for `cats/do-let`, so you can use any `cats` functions and macros that you'd use within a `do-let` within a `flow`".

PR #75, opened 2020-01-29
Helper for cljtest

caioaao:
PoC for discussion on `cljtest/match?`. This might not reduce boilerplate needed for bindings, but at least makes it more clear that binding results are being "unwrapped" and don't pollute the rest of the flow with unnecessary bindings.

### DIFF
--- a/src/state_flow/labs/cljtest.clj
+++ b/src/state_flow/labs/cljtest.clj
@@ -1,11 +1,18 @@
 (ns state-flow.labs.cljtest
   (:require [state-flow.core :as core]
             [state-flow.state :as state]
-            [clojure.test :as ctest]))
+            [clojure.test :as ctest]
+            matcher-combinators.clj-test))
+
+(defmacro testing-with {:style/indent :defn}
+  [desc bindings & body]
+  "Just like `testing`, but with bindings that are only accessible during test"
+  `(core/flow ~desc
+     [full-desc# (core/current-description)]
+     ~bindings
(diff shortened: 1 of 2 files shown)


sovelten:
I usually try to avoid adding more syntax surface whenever possible.
`match?` is only a macro because we need line numbering information.

You could have a similar behavior by using the following:
```clojure
(flow "test this thing"
  [actual bla]
  (expect expected actual))
```

See also https://github.com/nubank/state-flow/pull/71 for another idea (still coupled with matcher-combinators though)

I guess we'll also be thinking about a solution that does not rely on clojure.test as well.

caioaao:
I just don't like the idea of coupling a library that provides a way to structure tests with the assertions we use. Limiting users to `expect` doesn't make sense when we can have a more generic solution and leave the assertions to the user. We'll have to re-implement every use-case a user would need if we wen't down that road (`expect-throwing` and an 1-arity version that just checks if `actual` evaluates to true comes to mind already). That is, if we go down the route of relying on other testing frameworks.

sovelten:
This is just my personal opinion on macros, I'd rather use them very sparingly.

The point of `expect` is just to let you aware that this discussion is going on so you can voice your opinion.

We can merge this PR, though, if you like to experiment with it. I don't know if you want to merge it or just have a discussion about it.

Also another heads up is that we'll probably think of something that does not rely on clojure.test, at least not for building assertions.

caioaao:
No need to merge it. This is just a counterpoint to @philomates about boilerplate. And I also like the idea of not relying on clojure.test :ok_hand:

PR #76, opened 2020-02-03
refactor match?

dchelimsky:
- extract flow* function so flow and match? can both provide line number information
  - before this, match was not showing it's line number information
- make pop-meta a fn to align w/ push-meta
- align names and remove redundant defaults
- refactor cljtest_test

NOTE: this is extracted from #70, the functionality of which is still being discussed/evaluated. This PR just includes the refactoring (no changes to behaviour).

### DIFF
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -1,19 +1,14 @@
 (ns state-flow.cljtest
   (:require [cats.core :as m]
-            [clojure.test :as ctest :refer [is]]
+            [clojure.test :as t]
             [matcher-combinators.core :as matcher-combinators]
             [matcher-combinators.test]
             [state-flow.core :as core]
             [state-flow.probe :as probe]
             [state-flow.state :as state]))
 
-(defmacro match-expr
-  [desc value checker meta]
-  (with-meta
-    (list `ctest/testing desc (list `is (list 'match? checker value)))
-    meta))
-
 (defn match-probe
+  "Returns the right value returned by probe/probe."
   ([state matcher]
    (match-probe state matcher {}))
   ([state matcher params]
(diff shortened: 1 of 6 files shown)


dchelimsky:
Actually, this does include one enhancement - adding the line number info when reporting failed matches.

src/state_flow/core.clj
```diff
@@ -71,20 +75,25 @@
...
-(defmacro flow
-  "Defines a flow"
-  {:style/indent :defn}
-  [description & flows]
+(defn flow* [{:keys [description caller-meta]} & flows]
   (when-not (string-expr? description)
-     (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
-  (let [flow-meta (meta &form)
+    (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
+  (let [flow-meta caller-meta
```
sovelten: I really like the pattern of extracting unrelated macro code to a function, but I heard (don't remember where) it is not a good practice (or might lead to unexpected behavior) to use syntax quoting outside the context of a macro. Do you know anything about this?

src/state_flow/cljtest.clj
```diff
@@ -23,16 +18,24 @@
...
+  (core/flow*
+   {:description match-desc
+    :caller-meta (meta &form)}
+   `(m/do-let
+     [flow-desc# (core/current-description)
+      actual#    (if (state/state? ~actual)
+                   (match-probe ~actual ~expected ~params)
+                   (state/return ~actual))]
+     (state/wrap-fn #(t/testing flow-desc# (t/is (~'match? ~expected actual#))))
+     (state/return actual#))))
```
sovelten: looks much better :D

dchelimsky: Glad you like it.

src/state_flow/cljtest.clj
```diff
@@ -23,16 +18,24 @@
                         params))))
 
 (defmacro match?
-  "Builds a clojure.test assertion using matcher combinators"
-  [match-desc actual checker & [params]]
-  (let [form-meta (meta &form)]
```
sovelten: I'm puzzled as to why the previous approach didn't work. From my experience it does work sometimes (it was working when I did it the first time) and other times (actually most of the time) it doesn't.

src/state_flow/core.clj
dchelimsky: I haven't heard this, though I imagine it's because syntax quoting works differently within a macro than it does outside a macro (or, rather, the macro handles its result differently). In this case, because `do-let` is a macro, we can't use `(apply m/do-let flows')`, so we have to, at least, syntax quote that line to take advantage of splicing. If we only syntax quote `(m/do-let ~@flows')`, however, we end up with this

```clojure
(macroexpand-1 '(flow "desc"))
{:mfn #function[state-flow.state/reify--8727/fn--8733],
 :state-context #<State-E>}
```

As the code is in this PR, we get:

```
(macroexpand-1 '(flow "desc"))
(cats.core/do-let
 (state-flow.core/push-meta "desc" {:line 97, :column 17})
 [ret__18617__auto__ (cats.core/do-let (cats.core/return nil))]
 (state-flow.core/pop-meta)
 (cats.core/return ret__18617__auto__))
```

src/state_flow/cljtest.clj
dchelimsky: I never saw it work :)

I have made some other changes, so I'm not sure how far back to look, but if we look at master as of b90bb160b2e9bffc2c68a56507925f477b406d43, the `flow-desc` on https://github.com/nubank/state-flow/blob/b90bb160b2e9bffc2c68a56507925f477b406d43/src/state_flow/cljtest.clj#L30 does not include the line information from the `form-meta` here https://github.com/nubank/state-flow/blob/b90bb160b2e9bffc2c68a56507925f477b406d43/src/state_flow/cljtest.clj#L28, so it was never making it to the clojure.test output.

Issue #74, opened 2020-01-29
Uncaught exceptions are really hard debug


caioaao:
When an uncaught exception is thrown, all we get is the flow where it happened. Consider this flow and its execution output: https://gist.github.com/caioaao/0505aef68992f5bd6da88aba3e27a6f8

We don't have the step where it happened like we do with failed assertions. Instead we just have the flow name. This is specially hard when we have more complex flows that uses the faulty function more than once. Sub flows help mitigate this issue, but not completely solve it.

philomates:
Yes, I think addressing this would help a lot with the opaqueness of state-flow stacktraces

https://github.com/nubank/state-flow/pull/35 / https://github.com/nubank/state-flow/pull/51 do this for test failures, to some extent, but not exceptions.

I'm trying to figure out how to get line info for arbitrary steps in a flow. Here is what I have sketched but need to spend some more time with it, because I don't know if I'm on the right track

```clojure
(defmacro flow
  "Defines a flow"
  {:style/indent :defn}
  [description & flows]
  (when-not (string-expr? description)
     (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
  (let [flow-meta     (meta &form)
        flows'        (or flows `[(m/return nil)])
        subflow-lines (map (fn [f] (push-meta "" (meta f)))
                           flows')
        ;; add line meta info before each step in a flow body
        flows''       (interleave subflow-lines
                                  flows'
                                  (repeat pop-meta))]
    `(m/do-let
      (push-meta ~description ~flow-meta)
      [ret# (m/do-let ~@flows'')]
      pop-meta
      (m/return ret#))))
```

dchelimsky:
> We don't have the step where it happened like we do with failed assertions. Instead we just have the flow name. 

The stack trace [in the gist](https://gist.github.com/caioaao/0505aef68992f5bd6da88aba3e27a6f8) points directly to the problem, which is an NPE caused on line 6. I find that much more precise and useful than "line 13", from where the step was invoked.

caioaao:
Oh, yeah, the `line 6` is certainly important, but it's only a part of the
story. Not having the stacktrace to contextualize the error means
backtracking everytime there's an issue to find out where was the call that
broke it.

PR #77, opened 2020-02-03
Include line info for steps that raise exceptions

philomates:
An attempt to help address https://github.com/nubank/state-flow/issues/74

Given a flow that raises an exception, like:
```clojure
(ns state-flow.scratch
  (:require [state-flow.core :as state-flow]
            [state-flow.cljtest :refer [match?]]
            [state-flow.state :as state]))

(defn inc-a [s]
  (update s :a inc))

(defn change-state-buggy []
  (state/modify inc-a))

(state-flow/run!
  (state-flow/flow "buggy flow"
                   (change-state-buggy)
                   (match? "never run" 1 2)))
```

the result used to be like the following, which doesn't help pinpoint that `(change-state-buggy)` is the offending step:

```
                state-flow.state/reify/fn       state.clj:   63
                 state-flow.scratch/inc-a     scratch.clj:    7
                      clojure.core/update        core.clj: 6196
                         clojure.core/inc        core.clj:  927
                                      ...
java.lang.NullPointerException:

Syntax error (ExceptionInfo) compiling at (state_flow/scratch.clj:12:1).
Flow "buggy flow (scratch.clj:13)" failed with exception
```

With the changes in this PR:

```
                state-flow.state/reify/fn       state.clj:   63
                 state-flow.scratch/inc-a     scratch.clj:    7
                      clojure.core/update        core.clj: 6196
                         clojure.core/inc        core.clj:  927
                                      ...
java.lang.NullPointerException:

Syntax error (ExceptionInfo) compiling at (state_flow/scratch.clj:12:1).
Flow "buggy flow (scratch.clj:13) -> (change-state-buggy) (line 14)" failed with exception
```

**Note:**

we don't have line metadata info for variable references, so

```clojure
(state-flow/run!
  (state-flow/flow "buggy flow"
                   change-state-buggy))
```

results in 
```

Syntax error (ExceptionInfo) compiling at (state_flow/scratch.clj:12:1).
Flow "buggy flow (scratch.clj:13) -> change-state-buggy" failed with exception
```

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -71,14 +71,38 @@
   [s]
   (-> s meta :top-level-description))
 
+(def ^:private abbr-size 15)
+(defn ellipsify [expr-str]
+  (let [short-expr (subs expr-str 0 (- abbr-size 3))]
+    (case (first expr-str)
+      \( (str short-expr "...)")
+      \[ (str short-expr "...]")
+      (str short-expr "..."))))
+
+(defn abbr-sexpr [expr]
+  (let [expr-str   (str expr)
+        short-expr (if (< abbr-size (count expr-str))
+                     (ellipsify expr-str)
+                     expr-str)]
+    (str "`" short-expr "`")))
+
+(defn annote-with-line-meta [flows]
+  (let [subflow-lines (map (fn [f] `(push-meta ~(abbr-sexpr f)
+                                               ~(meta f)))
+                           flows)]
+    (interleave subflow-lines
+                flows
+                (repeat `pop-meta))))
+
 (defmacro flow
   "Defines a flow"
   {:style/indent :defn}
   [description & flows]
   (when-not (string-expr? description)
      (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
-  (let [flow-meta (meta &form)
-        flows'    (or flows `[(m/return nil)])]
+  (let [flow-meta     (meta &form)
+        flows'        (or (annote-with-line-meta flows)
+                          `[(m/return nil)])]
     `(m/do-let
       (push-meta ~description ~flow-meta)
       [ret# (m/do-let ~@flows')]
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -154,7 +154,7 @@
        (or (= 'flow (first expr))
            (= `flow (first expr)))))
 
-(defn annote-with-line-meta [flows]
+(defn annotate-with-line-meta [flows]
   (let [annotated-flows (reduce (fn [acc flow]
                                   (if (flow-expr? flow)
                                     (conj acc flow) ;; `flow`s push their own meta data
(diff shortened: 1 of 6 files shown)


src/state_flow/core.clj
```diff
@@ -75,14 +75,46 @@
...
   (let [flow-meta caller-meta
-        flows'    (or flows `[(m/return nil)])]
+        but-last-flows (butlast flows)
+        last-flow      (or (last flows)
+                           `(m/return nil))]
     `(m/do-let
       (push-meta ~description ~flow-meta)
-      [ret# (m/do-let ~@flows')]
+      ~@but-last-flows
+      ~(push-abbr-meta last-flow)
```
philomates: adding both the outer flow desc and the actual call sight info can be very useful. For example

```clojure
(state-flow/run!
  (state-flow/flow "buggy flow"




                   (change-state-buggy)))
```

correctly shows
```
Syntax error (ExceptionInfo) compiling at (state_flow/scratch.clj:12:1).
Flow "buggy flow (line 13) -> `(change-stat...)` (line 18)" failed with exception
```

sovelten:
You can also do `(flow "my-buggy-flow" (change-state-buggy))` whenever you need for debugging purposes and you would get correct line numbering then.

It seems like a better trade off than a complicated solution just to provide line number tracking for primitive steps.

caioaao:
This is a more contrived issue. I came across an interesting article about it, though, and might be a good direction to follow: https://pepeiborra.wordpress.com/2009/11/01/monadic-stack-traces-that-make-a-lot-of-sense/

I've only skimmed through the article, but the proposed solution is basically implementing an error monad that carries the stack trace

philomates:
I've cleaned the implementation and it should be ready to review again

src/state_flow/core.clj
```diff
@@ -121,6 +144,28 @@
...
+  `(push-meta ~(description/abbr-sexpr flow)
+              ~(assoc (meta flow)
+                      :call-site-meta true)))
+
+(defn- flow-expr? [expr]
+  (and (coll? expr)
+       (or (= 'flow (first expr))
+           (= `flow (first expr)))))
+
+(defn annote-with-line-meta [flows]
```
sovelten, line 157:
```suggestion
(defn annotate-with-line-meta [flows]
```

test/state_flow/core_test.clj
```diff
@@ -217,23 +209,24 @@
...
-                            level-1-line))))))
+          (is (<= line-number-before-flow-invocation
+                  level-3-line
+                  level-2-line
+                  level-1-line))))))
 
   (testing "after nested flows complete"
     (testing "within nested flows "
-      (is (re-matches #"level 1 \(core_test.clj:\d+\)"
+      (is (re-matches #"level 1 \(core_test.clj:\d+\) -> \(state-flow\/current-description\) \(line \d+\)"
```
gabrielgiussi, line 219: If I run this I get
```
=> "level 1 (core_test.clj:2) -> (state-flow/current-description) (line 4)"
```

Where `line 4` is coming from?

gabrielgiussi:
I couldn't find any test when a flow is throwing an exception and the line is included. I guess we are testing `current-description` instead and probably we use `current-description` when an exception is caught but then are not relaying on internal impl details? 🤔 

Shouldn't we have a test covering the case where an exception is raised and we have line information?

philomates:
Thanks for the suggestion @gabrielgiussi, updated with a test

Issue #72, opened 2020-01-27
`probe` sleeps `sleep-time` before the first try


dchelimsky:
Ideally, `probe` would always try the first time right away, and then sleep before subsequent tries, and right now there is no way to make that happen (besides setting `:sleep-time` to 0, but then it will try `:times-to-try` times with no sleep at all).

One solution would be to support a `:sleep-before-first-try` in `params`. The question is whether it should default to `true` or `false`. Hint: one of those is non-breaking, but the other is what I'd have expected/wanted in the first place.

dchelimsky:
Fixed in https://github.com/nubank/state-flow/pull/78

PR #78, opened 2020-02-04
probe refactoring

sovelten:
Refactoring of probe to allow users to gather more information from probe tries.

Tests are passing without change, so `probe` still has the same semantics, but also added `probe*` that returns a list of maps representing each try of `check-fn`.

This approach is (arguably) less ugly, more generic and also less efficient (because of non tail recursion) than the previous approach.

`sequence-while` could be moved to cats.core and made generic in order to work for any monad.

Needs polishing up and more tests, accepting feedback before going forward with this

### DIFF
--- a/src/state_flow/probe.clj
+++ b/src/state_flow/probe.clj
@@ -39,7 +39,7 @@
 (defn ^:private probe*
   "evaluates state repeatedly with check-fn until check-fn succeeds or we try too many times"
   ([flow check-fn]
-   (probe* flow check-fn {:sleep-time default-sleep-time :times-to-try default-times-to-try}))
+   (probe* flow check-fn {}))
   ([flow check-fn {:keys [sleep-time times-to-try]
                    :or   {sleep-time   default-sleep-time
                           times-to-try default-times-to-try}}]
--- a/src/state_flow/probe.clj
+++ b/src/state_flow/probe.clj
@@ -10,14 +10,14 @@
 (def default-times-to-try 5)
 
 (defn ^:private check
-  "Applies check-fn to return value of a flow, returns the check result and the original value in a map"
+  "Applies check-fn to the return value of a step, returns the check result and the original value in a map"
   [flow check-fn]
   (m/mlet [value flow]
     (state/return {:check-result (check-fn value)
                    :value        value})))
 
 (defn ^:private with-delay
-  "Adds a delay when the flow is run"
+  "Adds a delay before the step is run"
   [flow delay]
   (m/>> (state/wrap-fn #(Thread/sleep delay)) flow))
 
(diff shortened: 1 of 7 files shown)


src/state_flow/probe.clj
```diff
@@ -9,23 +9,51 @@
...
   ([state check-fn]
    (probe state check-fn {}))
-  ([state check-fn {:keys [sleep-time times-to-try]
-                    :or   {sleep-time   default-sleep-time
-                           times-to-try default-times-to-try}}]
-   (m/mlet [world (state/get)
-            :let [tries  (repeatedly #(do (Thread/sleep sleep-time) (state/eval state world)))
-                  result (retry times-to-try check-fn tries)]]
-     (state/return result))))
+  ([flow check-fn {:keys [sleep-time times-to-try]
```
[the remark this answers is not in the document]
sovelten: I agree with `step`, makes sense to me. Flows would be made of steps and bindings and a step can be a primitive step or another flow. Right?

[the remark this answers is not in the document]
dchelimsky: Yes! I'll submit a PR to update the readme and docs.

src/state_flow/probe.clj
```diff
@@ -9,23 +9,51 @@
...
+        (sequence-while* pred (state/return (conj results result)) (rest flows))))))
+
+(defn ^:private sequence-while
+  [pred flows]
+  (sequence-while* pred (state/return []) flows))
+
+(defn ^:private probe*
+  "evaluates state repeatedly with check-fn until check-fn succeeds or we try too many times"
+  ([flow check-fn]
+   (probe* flow check-fn {:sleep-time default-sleep-time :times-to-try default-times-to-try}))
```
dchelimsky: Don't need to declare the defaults here since they'll be assigned in the arity 3 version of this fn.

src/state_flow/probe.clj
```diff
@@ -9,23 +9,51 @@
...
+  ([flow check-fn {:keys [sleep-time times-to-try]
+                   :or   {sleep-time   default-sleep-time
+                          times-to-try default-times-to-try}}]
+   (sequence-while :check-result (repeat times-to-try (with-delay (check flow check-fn) sleep-time)))))
```
dchelimsky: While you're refactoring this, it would be great if the first try happened immediately, and the delay only happens after that. (https://github.com/nubank/state-flow/issues/72)

src/state_flow/probe.clj
```diff
@@ -9,23 +9,51 @@
...
-(defn ^:private retry
-  "Tries at most n times, returns a vector with true and first element that succeeded
-  or false and result of the first try"
-  [times-to-try check-fn tries]
-  (let [remaining (drop-while (complement check-fn) (take times-to-try tries))]
-    (if (empty? remaining)
-      [false (first tries)]
-      [true  (first remaining)])))
+(defn ^:private check
+  "Applies check-fn to return value of a flow, returns the check result and the original value in a map"
```
dchelimsky: s/flow/step

src/state_flow/probe.clj
```diff
@@ -9,23 +9,51 @@
...
+  [flow check-fn]
+  (m/mlet [value flow]
+    (state/return {:check-result (check-fn value)
+                   :value        value})))
+
+(defn ^:private with-delay
+  "Adds a delay when the flow is run"
```
dchelimsky: s/when/before

dchelimsky:
> Refactoring of probe to allow users to gather more information from probe tries.

What information are you talking about and how do you get to it?

sovelten:
> > Refactoring of probe to allow users to gather more information from probe tries.
> 
> What information are you talking about and how do you get to it?

`probe*` returns a list of maps, each map has the value being probed and the result of the check, so the user knows how many times were tried and the value and result each time. Before we only supplied the last value and the result of the last try, which is still the  behavior of `probe`

dchelimsky:
> probe* returns a list of maps, each map has the value being probed and the result of the check, so the user knows how many times were tried and the value and result each time. Before we only supplied the last value and the result of the last try, which is still the behavior of probe

In that case, shouldn't `probe*` be public?

sovelten:
> > probe* returns a list of maps, each map has the value being probed and the result of the check, so the user knows how many times were tried and the value and result each time. Before we only supplied the last value and the result of the last try, which is still the behavior of probe
> 
> In that case, shouldn't `probe*` be public?

Maybe. I wanted to replace `probe` with the new one actually. I'm also thinking whether this should  be exposed outside the library or just internal use. If it's internal use maybe we can keep it private but still use it internally. I'm accepting suggestions ;)

dchelimsky:
I wouldn't want to see it in flows directly, but at some point we may want to encourage ppl to write their own extensions for assertions and reporting. At that point, it might make sense for it to become public. I don't think that time is now, however.

sovelten:
> I wouldn't want to see it in flows directly, but at some point we may want to encourage ppl to write their own extensions for assertions and reporting. At that point, it might make sense for it to become public. I don't think that time is now, however.

Makes sense to me. I'll replace `probe` with the implementation of `probe*` and change the usages within the library. Do you thing `probe` should be private and accessed via `#'` or write a comment that this function is for internal use only?

dchelimsky:
> Do you thing probe should be private and accessed via #' or write a comment that this function is for internal use only?

I don't have a strong opinion about which is better, but I do think we should decide which approach to take and be consistent within the library.

dchelimsky:
@sovelten I see you've made some, but not all of the changes we discussed. What do you think of changing the `flow` param to `step`? And what about eliminating the delay before the first try?

sovelten:
@dchelimsky I eliminated the delay, hence the changes in the test results and some of the values to accomodate for the change. I changed the descriptions to use step, forgot to change the variables. Will do that!

dchelimsky:
> I eliminated the delay

Got it. Missed that line. Thanks!

PR #80, opened 2020-02-10
Introduce new match? fn in a new namespace and deprecate old one

dchelimsky:
RFC

This PR introduces `state-flow.assertions.matcher-combinators/match?` and deprecates `state-flow.cljtest/match?`.

### Rationale:
The existing `match?` fn takes `[description actual expected & [params]]`. 

In most cases, the `description` is redundant, so we're removing it from the API. If you want a description that is more informative than the default (you can see the expected value in the code and in the failure output), you can always wrap `match?` in a flow with a description.

The order (actual, then expected) is the reverse of clojure.test assertions and, consequently, the `match?` assert expression provided by matcher-combinators for use in clojure.test tests.

### Open questions

* Is `state-flow.assertions.xxx` a good idea?
  * yes, per discussion below
* Should we support `:description` in the `params` argument to `match?`? It works as of this PR in order to support the deprecated `cljtest/match?` fn with minimal duplication.
  * no - the refactoring tool gives the option to wrap it in a `flow` with a description
* As long as we're supporting the deprecated version, should we maintain a separate set of tests for it?

### TODO

* [x] add tools to migrate from old `match?` to new `match?`
* [ ] decouple `match?` from clojure.test (in a future PR)

### DIFF
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -5,15 +5,6 @@
             [state-flow.probe :as probe]
             [state-flow.state :as state]))
 
-(defn ^:private ensure-wrapped
-  "Internal use only.
-
-  Given a state-flow step, returns value as/is, else wraps value in a state-flow step."
-  [value]
-  (if (state/state? value)
-    value
-    (state/return value)))
-
 (defn ^:private match-probe
   "Internal use only.
 
@@ -57,8 +48,8 @@
       `(m/do-let
         [flow-desc# (core/current-description)
          actual#    (if (> (:times-to-try ~params*) 1)
-                      (#'match-probe (#'ensure-wrapped ~actual) ~expected ~params*)
-                      (#'ensure-wrapped ~actual))]
+                      (#'match-probe (state/ensure-step ~actual) ~expected ~params*)
+                      (state/ensure-step ~actual))]
         ;; TODO: (dchelimsky, 2020-02-11) we plan to decouple
         ;; assertions from reporting in a future release. Remove this
         ;; next line when that happens.
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -37,17 +37,16 @@
   retry :times-to-try times with :sleep-time
 
   See `state-flow.probe/probe` for more info"
-  ([expected actual]
-   `(match? ~expected ~actual {:times-to-try 1}))
-  ([expected actual {:keys [times-to-try]
-                     :or {times-to-try 1}
-                     :as params}]
+  [expected actual & [{:keys [times-to-try]
+                       :or {times-to-try 1}
+                       :as params}]]
    ;; description is here to support the
    ;; deprecated cljtest/match? fn.  Undecided
    ;; whether we want to make it part of the API.
    ;; caller-meta is definitely not part of the API.
    (let [params* (merge {:description "match?"
-                         :caller-meta (meta &form)}
+                         :caller-meta (meta &form)
+                         :times-to-try 1}
                        params)]
      (core/flow*
       {:description (:description params*)
@@ -60,5 +59,15 @@
          actual#    (if (> (:times-to-try ~params*) 1)
                       (#'match-probe (#'ensure-wrapped ~actual) ~expected ~params*)
                       (#'ensure-wrapped ~actual))]
-        (state/wrap-fn #(t/testing flow-desc# (t/is (~'match? ~expected actual#))))
-        (state/return actual#))))))
+        ;; TODO: (dchelimsky, 2020-02-11) we plan to decouple
+        ;; assertions from reporting in a future release. Remove this
+        ;; next line when that happens.
+        ;; NOTE: the match? symbol on this next line is used to
+        ;; dispatch clojure.test's assert-expr multimethod to an
+        ;; implementation in matcher-combinators, and there is no way
+        ;; to qualify it with a namespace. This means that if you're
+        ;; exploring this macro in _this_ namespace, and try to invoke
+        ;; it, you'll get a stack overflow because the compiler
+        ;; interprets match? as this macro.
+        (state/wrap-fn #(~'clojure.test/testing flow-desc# (~'clojure.test/is (~'match? ~expected actual#))))
+        (state/return actual#)))))
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -1,0 +1,57 @@
+(ns state-flow.assertions.matcher-combinators
+  (:require [cats.core :as m]
+            [matcher-combinators.core :as matcher-combinators]
+            [state-flow.core :as core]
+            [state-flow.probe :as probe]
+            [state-flow.state :as state]))
+
+(defn ^:private ensure-wrapped
+  "Internal use only.
+
+  Given a state-flow step, returns value as/is, else wraps value in a state-flow step."
+  [value]
+  (if (state/state? value)
+    value
+    (state/return value)))
+
+(defn ^:private match-probe
+  "Internal use only.
+
+  Returns the right value returned by probe/probe."
+  [state matcher params]
+  (m/fmap second
+          (probe/probe state
+                       #(matcher-combinators/match? (matcher-combinators/match matcher %))
+                       params)))
+
+(defmacro match?
+  "Builds a state-flow assertion using matcher-combinators.
+
+  - expected can be a literal value or a matcher-combinators matcher
+  - actual can be a literal value, a primitive, or a flow
+  - params is an optional map supporting:
+    - :times-to-try optional, default 1
+    - :sleep-time   optional, default 200
+
+  Given (> :times-to-try 1), match? will use `state-flow-probe/probe` to
+  retry :times-to-try times with :sleep-time
+
+  See `state-flow.probe/probe` for more info"
+  ([expected actual]
+   `(match? ~expected ~actual {:times-to-try 1}))
+  ([expected actual {:keys [times-to-try]
+                     :or {times-to-try 1}
+                     :as params}]
+   (core/flow*
+    {:description "match?"
+     :caller-meta (meta &form)}
+    ;; Nesting m/do-let inside a call the function core/flow* is
+    ;; a bit ugly, but it supports getting the correct line number
+    ;; information from core/current-description.
+    `(m/do-let
+      [flow-desc# (core/current-description)
+       actual#    (if (> (:times-to-try ~params) 1)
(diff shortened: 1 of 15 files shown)


src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -0,0 +1,64 @@
...
+       :caller-meta (:caller-meta params*)}
+      ;; Nesting m/do-let inside a call the function core/flow* is
+      ;; a bit ugly, but it supports getting the correct line number
+      ;; information from core/current-description.
+      `(m/do-let
+        [flow-desc# (core/current-description)
+         actual#    (if (> (:times-to-try ~params*) 1)
+                      (#'match-probe (#'ensure-wrapped ~actual) ~expected ~params*)
+                      (#'ensure-wrapped ~actual))]
+        (state/wrap-fn #(t/testing flow-desc# (t/is (~'match? ~expected actual#))))
```
sovelten: We should stop using clojure.test for assertions in this new `match?`, right?

dchelimsky:
> We should stop using clojure.test for assertions in this new match?, right?

Yes, but that's an implementation detail that will require more thought/design, and I want to get this `match?` working and out sooner than later so we can roll out changes to playbook, etc. Agree w/ the approach?

sovelten:
My biggest concert is the alternative to the missing description.

I wonder if:
```clojure
(match? "this and that" actual expected)
```
is well replaced by:
```clojure
(flow "this and that" (match? expected actual))
```
At least the refactoring should be possible to do automatically.

Should we provide another fn with description?

dchelimsky:
I think `(flow "this and that" (match? expected actual))` is the best solution for when you want a description that provides more information than the expected and actual values do. And yes, we can definitely have the automated refactoring do that, although perhaps that could be a configurable option e.g. "Preserve descriptions? yes/no". I really do want to discourage redundant and/or useless descriptions, of which there are many, at least in the tests for state-flow and state-flow helpers.

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -0,0 +1,64 @@
...
+     (core/flow*
+      {:description (:description params*)
+       :caller-meta (:caller-meta params*)}
+      ;; Nesting m/do-let inside a call the function core/flow* is
+      ;; a bit ugly, but it supports getting the correct line number
+      ;; information from core/current-description.
+      `(m/do-let
+        [flow-desc# (core/current-description)
+         actual#    (if (> (:times-to-try ~params*) 1)
+                      (#'match-probe (#'ensure-wrapped ~actual) ~expected ~params*)
```
sovelten: Another issue is about probing. Do we want probe by default? Or maybe provide a `match-probe?` for doing it explicitly?

dchelimsky:
> Another issue is about probing. Do we want probe by default? Or maybe provide a match-probe? for doing it explicitly?

I didn't document this well, but I think probe should only happen when you explicitly set `:times-to-try` to something > 1. That's in the implementation, but not documented (I'll add that now).

sovelten:
> I didn't document this well, but I think probe should only happen when you explicitly set `:times-to-try` to something > 1. That's in the implementation, but not documented (I'll add that now).

I don't have a strong opinion either way, but if we accept only values and not steps, this might lead to a more uniform syntax and possibly less confusion. The only reason that we accept values or steps on _actual_ is because of probing. So without accepting steps as arguments we are required to do something of the following:

```clojure
(flow "my test"
  [actual get-value-step]
  (match? expected actual))
```
And when we explicitly want to probe:
```clojure
(flow "my test" (match-probe expected get-value-step))
```

dchelimsky:
Hmm. Well, I <3ed that right away and then realized that `(match? expected get-value-step)` might still be preferable. I'm going to slow down and think about it a bit longer :)

fernando-nubank:
> I didn't document this well, but I think probe should only happen when you explicitly set :times-to-try to something > 1. That's in the implementation, but not documented (I'll add that now)

Please make this probe behavior explicit - sometimes this is very annoying when you are still writing an integration test and this slows down the feedback.

dchelimsky:
> Please make this probe behavior explicit - sometimes this is very annoying when you are still writing an integration test and this slows down the feedback.

One way or the other it will be. The way I implemented it, you have to include a params argument with `{:times-to-try [something > 1]}`. The way @sovelten is recommending would also be explicit.

fernando-nubank:
Oh, got it. Liked a lot this API 
```clojure
(flow "my test" (match-probe expected get-value-step))
```
but, for me, both ideas work as they are actually explicit

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -0,0 +1,64 @@
+(ns state-flow.assertions.matcher-combinators
```
philomates: `matcher-combinators` is also test framework agnostic, so I would only use this ns name if it doesn't include the use of `clojure.test` stuffs

dchelimsky: We need to decouple reporting in order to do that, and we're not there yet. Nobody will feel any pain if we release this as/is and decouple reporting later, whereas the longer we wait to get this change out the more people will have to migrate their state-flow suites later. How about I just add a comment that we'll be decoupling that later right in the code?

philomates: ah, makes sense, sounds good to me

dchelimsky: @philomates do you like the idea of grouping assertions under`state-flow.assertions.xxx`?

philomates: Yup; and then folks that wanted midje or expectations, or vanilla clojure.test could load related `state-flow.assertions.xxx` namespaces, right?

dchelimsky: That's the idea. In theory we could move `testing` there, and add an `is`, etc. We'll have to do that very carefully - consider that `is` already has a `match?` assert-expr (if you load matcher-combinators), etc. That could get confusing.

dchelimsky:
> I don't have a strong opinion either way, but if we accept only values and not steps, this might lead to a more uniform syntax and possibly less confusion.

@sovelten I want to decouple the value/step choice from retry. I think it's perfectly valid for `match?` to receive a step and _not_ retry. If we only accept values for `match?` and only accept steps for `match-probe`, then we're disallowing that case (or allowing it with a specific `(match-probe expected actual {:times-to-try 1})`, which I find awkward.

All of this to say, I'm leaning more toward what I've already implemented:

```clojure
(match? expected-value-or-matcher actual-value-or-step)
(match? expected-value-or-matcher actual-value-or-step {:times-to-try 3})
```

Both support all combinations of expected/actual, but only the latter retries. I think the docstring explains this clearly, and would be happy to improve it if you think something else would clarify it further. WDYT?

sovelten:
> @sovelten I want to decouple the value/step choice from retry. I think it's perfectly valid for `match?` to receive a step and _not_ retry. If we only accept values for `match?` and only accept steps for `match-probe`, then we're disallowing that case (or allowing it with a specific `(match-probe expected actual {:times-to-try 1})`, which I find awkward.
> 
> All of this to say, I'm leaning more toward what I've already implemented:
> 
> ```clojure
> (match? expected-value-or-matcher actual-value-or-step)
> (match? expected-value-or-matcher actual-value-or-step {:times-to-try 3})
> ```
> 
> Both support all combinations of expected/actual, but only the latter retries. I think the docstring explains this clearly, and would be happy to improve it if you think something else would clarify it further. WDYT?

There are trade offs, I don't see a clear winner, but I'm fine with the solution.

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -0,0 +1,73 @@
...
+            [state-flow.state :as state]))
+
+(defn ^:private ensure-wrapped
+  "Internal use only.
+
+  Given a state-flow step, returns value as/is, else wraps value in a state-flow step."
+  [value]
+  (if (state/state? value)
+    value
+    (state/return value)))
```
sovelten: This looks general enough to be on `state.clj`, WDYT?

dchelimsky: Agreed it's general, but I don't necessarily want to promote it to a public API. You good with "For internal use. Subject to change." in the docstring?

test/state_flow/cljtest_test.clj
```diff
@@ -2,6 +2,7 @@
   (:require [clojure.test :as t :refer [deftest testing is]]
             [matcher-combinators.test :refer [match?]]
             [matcher-combinators.matchers :as matchers]
+            [state-flow.assertions.matcher-combinators :as amc]
```
sovelten: We usually avoid short acronyms like this. As per Stuart Sierra namespace alias guidelines which are always sensible and good: https://stuartsierra.com/2015/05/10/clojure-namespace-aliases

dchelimsky: Yeah, I usually recommend reading that, myself. But I succumbed to the "I don't want to type all those letters" voice! Will fix.

src/state_flow/assertions/matcher_combinators.clj
sovelten: Yes, that works

test/state_flow/cljtest_test.clj
dchelimsky: ✅

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: ✅

dchelimsky:
@sovelten @philomates I've added a function to help refactor from the old match to the new match. I think this is ready to go at this point. Care to take one last look before I merge?

dchelimsky:
Actually, I just had one more thought - we should offer a refactoring option (you'd have to opt-in) to ensure the default params stay the same as before, e.g.

```clojure
(match? desc actual expected)
;; becomes
(match? expected actual {:times-to-try 5
                         :sleep-time 200})
```

If the source match? already has either or both param defined, that would take precedence (i.e. merge the orig map into the default map)

This would result in behavioral consistency, and also make it obvious to the developer when the params are being used unnecessarily.

WDYT?

sovelten:
> Actually, I just had one more thought - we should offer a refactoring option (you'd have to opt-in) to ensure the default params stay the same as before, e.g.
> 
> ```clojure
> (match? desc actual expected)
> ;; becomes
> (match? expected actual {:times-to-try 5
>                          :sleep-time 200})
> ```
> 
> If the source match? already has either or both param defined, that would take precedence (i.e. merge the orig map into the default map)
> 
> This would result in behavioral consistency, and also make it obvious to the developer when the params are being used unnecessarily.
> 
> WDYT?

Very few cases need to actually do probing, but I think it makes sense to add this option. It's kind of a nice to have though.

PR #81, opened 2020-02-11
Use vectors instead of pairs

sovelten:
cats Pair data structure is not very idiomatic and vector is perfectly fine and less confusing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [2.2.1]
+- Use vectors in internal state data structure instead of cats pairs
+
 ## [2.2.0]
 
 - Remove delay from the first try of `probe`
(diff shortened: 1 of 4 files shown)


dchelimsky:
This looks great! I'm going to test it out locally, and we should update the README and walkthrough to eliminate references to Pair. If it's OK w/ you, @sovelten, I'll add a commit with the doc changes to this PR. Otherwise I could do it in a follow up PR.

dchelimsky:
I updated the walkthrough in https://github.com/nubank/state-flow/commit/325eaa9ede08d2da4fe3b8f94c4243ce75790917, which is on the https://github.com/nubank/state-flow/commits/dc/use-vectors-docs branch (branched off this PR's branch). @sovelten please feel free to merge that commit into this branch to include in this PR. If you don't, I'll submit a PR afterward.

PR #86, opened 2020-02-14
Introduce new match? fn in a new namespace and deprecate old one (take 2)

dchelimsky:
This PR reintroduces the changes from PR #80, but fixes an issue that caused `match?` to fail to work.

The problem was that the matcher-combinators.test/match? assert-expr was never registered because that namespace was not being required. We didn't see it in development because it _is_ required from tests.

### DIFF
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -1,6 +1,7 @@
 (ns state-flow.assertions.matcher-combinators
   (:require [cats.core :as m]
             [matcher-combinators.core :as matcher-combinators]
+            [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
             [state-flow.core :as core]
             [state-flow.probe :as probe]
             [state-flow.state :as state]))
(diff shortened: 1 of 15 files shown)


dchelimsky:
PS - I verified this by installing 2.2.4 locally and using it from another project. All OK.

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -0,0 +1,58 @@
+(ns state-flow.assertions.matcher-combinators
+  (:require [cats.core :as m]
+            [matcher-combinators.core :as matcher-combinators]
+            [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
```
philomates: I believe you could also explicitly do a `:refer [match?]`

dchelimsky: Not in this case, because we define a `match?` var in this namespace (so you'd get "match? already refers ...). It works in namespaces that don't def a `match?` var, like most test namespaces.

PR #89, opened 2020-02-17
Create release automation

fernando-nubank:
This adds a Github Action that publishes new `state-flow` versions to Clojars. 
For more info please check RELEASING.md

:warning: One downside of this automation is that we will not sign new releases with GPG.

### Before merging
 - [x] Create a Clojars robot user
 - [x] Add user credentials as Github Secrets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: Clojars
+
+on:
+  push:
+    tags:
+    - '*'
+
+jobs:
+  release:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: lein deps
+
+    - name: Run tests
+      run: lein test
+
+    - name: Publish
+      run: lein deploy publish
+      env:
+          CLOJARS_USERNAME: ${{ secrets.CLOJARS_USERNAME }}
+          CLOJARS_PASSWD: ${{ secrets.CLOJARS_PASSWD }}
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,16 @@
+# Releasing
+
+To release a new version in Clojars, you need to run 
+
+```bash
+./release.sh
+```
+
+This command creates a git tag with the project's current version and pushes it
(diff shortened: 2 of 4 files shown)


markomafs:
We could work in signed commits in the future.

sovelten:
@fernando-nubank That's a major downside, no? I thought it was not even possible to submit unsigned packages to Clojars

dchelimsky:
>> One downside of this automation is that we will not sign new releases with GPG.

> @fernando-nubank That's a major downside, no? I thought it was not even possible to submit unsigned packages to Clojars

Interestingly, I had to get my gpg setup working with lein in order to deploy 2.2.4. Then I deleted it from my local repo and re-installed it from clojars. Then:

```
> jarsigner -verify ~/.m2/repository/nubank/state-flow/2.2.4/state-flow-2.2.4.jar

jar is unsigned.
```

¯\\_(ツ)_/¯

fernando-nubank:
> That's a major downside, no? I thought it was not even possible to submit unsigned packages to Clojars

Not from what we are doing now, I guess. Do we have a good key to sign the releases? I mean, one that has a good reputation in the community?

sovelten:
> Not from what we are doing now, I guess. Do we have a good key to sign the releases? I mean, one that has a good reputation in the community?

No, but shouldn't we have one for Nubank or something? I don't really know much about this to be honest, just asking out of curiosity.

fernando-nubank:
> No, but shouldn't we have one for Nubank or something? I don't really know much about this to be honest, just asking out of curiosity.

Yeah, I think we should. But this will be a lot of work, I guess. For now, I'm ok with releasing unsigned.

fernando-nubank:
@sovelten @dchelimsky Can I merge this?

dchelimsky:
Looking at the clojars repo, I see that there are signed jars there: https://repo.clojars.org/nubank/state-flow/2.1.0/ (.asc files). What I'm not clear about is why the jars we download are not the signed ones.

re: the value of signed jars, here's an interesting discussion: https://clojureverse.org/t/should-you-sign-jars-you-publish-on-clojars/4093

If we're OK w/ unsigned artifacts, then yes, we can merge this.

PR #91, opened 2020-02-20
Fix exception on fmap

caioaao:
Previously the added testcase would __throw__:

```clojure
java.lang.ClassCastException: cats.monad.exception.Failure cannot be cast to java.lang.Number
```

### DIFF
--- a/src/state_flow/state.clj
+++ b/src/state_flow/state.clj
@@ -9,7 +9,7 @@
 
 (declare error-context)
 
-(defn- result-or-err [f args]
+(defn- result-or-err [f & args]
   (let [result ((e/wrap (partial apply f)) args)]
     (if (e/failure? result)
       result
@@ -49,7 +49,7 @@
 
     p/Monad
     (-mreturn [_ v]
-      (error-state (partial vector v)))
+      (error-state #(vector v %)))
 
     (-mbind [_ self f]
       (error-state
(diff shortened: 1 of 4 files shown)


sovelten:
good catch!
can you add changelog and update version?

src/state_flow/state.clj
```diff
@@ -21,8 +20,17 @@
     p/Functor
     (-fmap [_ f fv]
       (state/state (fn [s]
-                     (let [[v ns]  ((p/-extract fv) s)]
-                       [(f v) ns]))
+                     (let [mp ((e/wrap (p/-extract fv)) s)]
+                       (cond
+                         (e/failure? mp)
```
dchelimsky: Only one of these cases is tested. Are both actually necessary?

caioaao:
Turns out the issue was more widespread. I've added testcases to cover all issues I found and refactored the error state monad to handle them appropriately.

caioaao:
tests are green :heavy_check_mark:

test/state_flow/state_test.clj
```diff
@@ -11,7 +10,24 @@
     (is (state/gets inc))
```
dchelimsky: These can be removed now that all of them are being used in tests below.

sovelten:
What's the benefit of using a new Record to encapsulate state? Is it for being able to distinguish from the vanilla implementation from cats?

caioaao:
@sovelten it's to implement the correct `p/-extract`

PR #97, opened 2020-02-27
handle anon fns in refactor_match script

dchelimsky:
* handle anonymous functions
  * preserve map structures
* make script depend on state-flow repo instead of a jar

### DIFF
--- a/bin/refactor-match.sh
+++ b/bin/refactor-match.sh
@@ -1,9 +1,10 @@
 #!/bin/sh
 #_(
-   DEPS='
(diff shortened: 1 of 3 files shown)


dchelimsky:
> I don't get what was the change to avoid map reordering

It was doing the reordering (and adding commas) when converting to an sexpr, which the fix I had on the branch earlier was doing for everything, not just functions. In this PR, functions get `sexpr`d, but everything else gets `node`d. Make sense?

PR #100, opened 2020-03-02
add section on writing helpers to README

dchelimsky:


README.md
```diff
@@ -284,3 +284,62 @@ and we also have a step that fetches this data from db (`fetch-data`). We want t
...
+            (map :user/first-name users)))
+```
+
+The `http/request` helper could be implemented as simply as this:
+
+``` clojure
+(ns http-helpers
+  (:require [my-app.http :as http))
+(defn request [req]
+  (state-flow.state/wrap-fn #(http/request req))
```
sovelten: `wrap-fn` is actually to be used as a workaround when there is no other option.
Ideally your state has a mocked http component that is able to make requests, you get the component from the state and do something with it.

README.md
```diff
@@ -284,3 +284,62 @@ and we also have a step that fetches this data from db (`fetch-data`). We want t
...
+```
+
+This produces a step that can be used in a flow, as above.
+
+### funcool.cats
```
sovelten: This section looks great!

README.md
dchelimsky: @sovelten updated to use `state/return` with a `binding`. Better?

PR #103, opened 2020-03-04
update readme with `run!` deprecation and `match?` namespace change

philomates:


README.md
```diff
@@ -19,7 +19,7 @@ reference. Use the `flow` macro to define a flow:
 (flow <description> <step/bindings>*)
 ```
 
-Once defined, you can run it with `(state-flow.core/run! (flow ...) <initial-state>)`.
+Once defined, you can run it with `(state-flow.core/run* (flow ...) <initial-state>)`.
```
dchelimsky: This is wrong. `state-flow.core/run*` takes an options map first, followed by the flow.

README.md
```diff
@@ -35,8 +35,8 @@ a system using [Stuart Sierra's Component](https://github.com/stuartsierra/compo
 ```clojure
 (def a-flow (flow ...))
 
-(state-flow.core/run! flow <one-initial-state>)
-(state-flow.core/run! flow <another-initial-state>)
+(state-flow.core/run* flow <one-initial-state>)
```
dchelimsky: Same problem here ....

README.md
dchelimsky: This is actually a concern for me. I think the differences between `run` and `run*` are confusing (as evidenced by this PR). Ideally we'd have signatures that are more similar like `(run flow init-state)` and `(run* flow opts)`, but that would be hard to do at this point. So a new name would probably be in order to address this. `exec` and `eval` have specific (arbitrary IMO) meaning in cats, so we can't really use those. Other ideas?

dchelimsky: Alternatively, we undeprecate `run!` and leave `run*` in place with guidance to use it for helpers, etc.

Issue #108, opened 2020-04-06
The match? assertion should return either a boolean or other indication of success/failure


dchelimsky:
The `state-flow.assertions.matcher-combinators/match?` step (monad) returns a value:

```clojure
(run
  (flow "increment count"
    (state/modify update :count inc)
    (match? 1 (state/gets :count)))
  {:count 0})
;; => [1 {:count 1}]
```

The left value, 1, doesn't help us understand whether the `match?` succeeded or failed. Do know that, we have to look at the REPL or the output log.

I recommend we change it to return a map with the success/fail information matcher-combinators reports to clojure test.

sovelten:
`match?` should actually return the return value of the right value (now that the order is inverted).

This is important for probing. If you need to do a probe test and you also need to use whatever the probed step returns.

```clojure
...
[actual (match? expected probe-step {:times 2})]
(some-request actual)
...
```

If it wasn't for that you could do
```clojure
[actual step
 test-result (match? expected actual)]
```

Probably one more reason for splitting match? and mach-probe?

Another solution is to store both in a map `{:expected ... :actual ... :report ...}`

```clojure
...
[{the-thing-i-need :actual} (match? expected probe-step {:times 2})]
(some-request the-thing-i-need)
...
```

dchelimsky:
`match?` uses `probe` internally, and can continue to do while still returning something different. Why would you do this:

```clojure
...
[actual (match? expected probe-step {:times 2})]
(some-request actual)
...
```

When you could just use probe directly, if that's your intent.

dchelimsky:
re: `{:expected ... :actual ... :report ...}`, I like that idea, though I think that should all be in the report returned from matcher-combinators.

sovelten:
> `match?` uses `probe` internally, and can continue to do while still returning something different. Why would you do this:
> 
> ```clojure
> ...
> [actual (match? expected probe-step {:times 2})]
> (some-request actual)
> ...
> ```
> 
> When you could just use probe directly, if that's your intent.

You want to make a match? assertion and you also want the return value of the probe-step. Does that make sense? `probe` is set for internal use only right now.

dchelimsky:
> breaking change though

I realize it's a change in "how it works" if you happen to know how `match?` works today, but it's return is not documented, so for me it's not a breaking change. Also, I'm going to guess that you're the only user using it this way :)

> You want to make a match? assertion and you also want the return value of the probe-step.

[...]

PR #110, opened 2020-04-06
return a map from match?

dchelimsky:


dchelimsky:
Addressing #108

dchelimsky:
@sovelten I still need to add the `report->actual` fn before I submit this as "not a draft". Feel free to review now or wait.

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -53,9 +63,13 @@
        [flow-desc# (core/current-description)
         actual#    (if (> (:times-to-try ~params*) 1)
                      (#'match-probe (state/ensure-step ~actual) ~expected ~params*)
-                     (state/ensure-step ~actual))]
+                     (state/ensure-step ~actual))
+        report#    (state/return (matcher-combinators/match ~expected actual#))]
+       (state/modify update :match/results (fnil conj []) (assoc report# :match/desc flow-desc#))
```
sovelten, line 68: Are we going to modify the state or the meta?
I was thinking that state-flow specific stuff could be on the meta, while user specific stuff on the state, so as not to conflict with whatever the user wants to have as a state.

sovelten: On another note, this statement is a bit too big, maybe it deserves its own function

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -9,31 +9,41 @@
...
-  [state matcher params]
+  Returns the right value returned by probe/probe.
+
+  args:
+  - step is a step (monad) which should produce the actual value
+  - matcher is a matcher-combinators matcher object
+  - params are passed directly to probe "
+  [step matcher params]
   (m/fmap (comp :value last)
-          (probe/probe state
```
sovelten, line 15: You want to fetch the result of all tries instead of only value of the last try here.

dchelimsky: Roger.

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: Side note: this doesn't even need to be here yet. I added it (yes, to the wrong place - will put on meta) with reporting in mind, so the runner can analyze the match results and print failure information when appropriate. I may end up removing it from here for now and restore it when we get to that decoupling.

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: @sovelten I removed the bit about storing these results in the state (or meta) from this PR, so I don't think that matters (for this PR). Agree?

dchelimsky:
@sovelten I added the `report->actual` fn we discussed in #108. I'm not in love with the name, but I don't know that I have a better choice. WDYT about `extract-actual`, or something cute like `<<actual`?

```clojure
[actual (<<actual (match? 3 (+ 1 2)))]
```

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: I changed my mind. The return of match? now looks like this when (> times-to-try 1)

```clojure
{:probe/sleep-time   110
 :probe/times-to-try 3
 :probe/results      [{:check-result false :value 0}
                      {:check-result true  :value 2}]
 :match/expected     2
 :match/actual       2
 :match/result       :match}
```

dchelimsky: FYI - adding this to the meta of the state will allow us to move the cljtest reporting to the cljtest namespace

sovelten: Looks good to me :tada:

dchelimsky: One thing I'm not sure about is whether we should document the keys in that map or just say "a map with information about the success or failure of the match, the details of which are used internally by state-flow and subject to change". I want to discourage anybody, for now, from treating that map's shape as an API the you can program against. I might see this differently once we have the decoupling design worked out.

sovelten: I agree, I would keep this model as internal for now (or ever I don't know)

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -51,11 +64,36 @@
...
+       actual (report->actual report)]
+      (state/return actual))
+
+    (flow \"example\"
+      [actual (report->actual (match? 2 3))]
+      (state/return actual)) "
+  [report-or-match]
+  `(m/do-let
+    [report# (state/ensure-step ~report-or-match)]
+    (state/return (:match/actual report#))))
```
sovelten, line 99: I'm not a fan of this approach. I'm not a fan of the so called "true union types" like things can be one thing or the other.
I find it quite confusing, not well-typed :grimacing: .

What are the problems you see with just a function? Is it to avoid exposure to fmap? fmap is beautiful :sparkles: . But also avoidable with just a function by using `:let`.

Also I don't understand the need for a macro here.

dchelimsky: `match?` supports a step or a value for `actual`. I was just going along w/ that idea, but I appreciate your perspective here. Changing it to a function.

sovelten: Yeah, maybe it is just a longing for orderliness but that is also one reason why I was defending a different match? for probing.

But that may be a different case, because you cannot write match-probe? in terms of match? but you can compose `report->actual` with fmap.

Just rambling, don't mind me.

dchelimsky: OK @sovelten - it's a fn now and there's even `fmap` in the test.

test/state_flow/cljtest_test.clj
```diff
@@ -1,112 +1,42 @@
 (ns state-flow.cljtest-test
   (:require [clojure.test :as t :refer [deftest testing is]]
             [matcher-combinators.test :refer [match?]]
-            [matcher-combinators.matchers :as matchers]
-            [state-flow.assertions.matcher-combinators :as assertions.matcher-combinators]
-            [state-flow.test-helpers :as test-helpers :refer [this-line-number]]
+            [state-flow.assertions.matcher-combinators :as mc]
```
sovelten, line 4: avoid acronyms?

dchelimsky: Totally agree 99.9999% of the time. `assertions.matcher-combinators/xxx` all over the place was making my eyes bleed.

test/state_flow/test_helpers.clj
```diff
@@ -14,28 +14,38 @@
...
-(defn delayed-add-two
-  [delay-ms]
-  "Changes state in the future"
+(defn delayed-modify
+  "Applies f to state + any additional args in the future."
+  [delay-ms f & args]
   (state/modify (fn [state]
                   (future (do (Thread/sleep delay-ms)
-                              (swap! (:value state) + 2)))
+                              (apply f state args)))
```
sovelten, line 22: That only works for atoms though. Applying a pure function would be kinda pointless.

dchelimsky: Maybe make all of state an atom? Then we could have `(apply swap! state f args)`

test/state_flow/cljtest_test.clj
dchelimsky: Also, `matcher-combinators/xxx` would be confusing, since it's not coming from the matcher-combinators library. I guess `assertions` could work, but that's also confusing :) So I went with a plua (you know, a poor, lonely, unwanted acronym).

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -51,11 +61,22 @@
...
-        ;; next line when that happens.
+        probe-res# (#'match-probe (state/ensure-step ~actual) ~expected ~params*)
+        :let [actual# (-> probe-res# last :value)
+              report# (cond-> (assoc (matcher-combinators/match ~expected actual#)
+                                     :match/expected ~expected
+                                     :match/actual   actual#)
+                        (> (count probe-res#) 1)
+                        (assoc :probe/results      probe-res#
+                               :probe/sleep-time   ~(:sleep-time params*)
+                               :probe/times-to-try ~(:times-to-try params*)))]]
```
sovelten, line 72: I suggest replacing `cond->`
```clojure
(merge (assoc ....)
  (when (> (count probe-res#) 1)
    {:probe/restuls probre-res# ...}))
```

I think it is more readable than `cond->`

test/state_flow/cljtest_test.clj
sovelten: how about `matcher`?

test/state_flow/test_helpers.clj
sovelten: I think we expect state to be a map. If the whole state is an atom there is no point of passing it around, the atom is always the same. But this is test code, I'm not really concerned. Just wondering why you took that route.

dchelimsky: I find that every time I look at these tests I have to go look at the definition of `delayed-add-two`, so I wanted to refactor it so I can see what's happening in the test while still avoiding some of the duplication.

test/state_flow/cljtest_test.clj
dchelimsky: I'd rather just go back to `assertions.matcher-combinators`, which won't be confused with things in other parts of the test suite.

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: You say patota, I say potato .... (inside joke between me and me)

I just left out the conditional. Objectively cleaner than either `cond->` or `merge .. when`, and why not keep all the info?

test/state_flow/cljtest_test.clj
dchelimsky: I could just refer `match?` and `report->actual`, but then we'd end up with:

```clojure
(testing "with explicit matcher for expected"
  (let [[ret state] (state/run (match? (matchers/equals 3) 3) {})]
    (is (match? {:match/result :match} ret))))
```

The two `match?`s are different. They both work because one is a function and the other is a symbol used to dispatch a multimethod, but they are confusing together!

dchelimsky: I'm going with the acronym for this PR. We can always revisit this later.

CHANGELOG.md
```diff
@@ -1,7 +1,21 @@
...
-
+## [3.0.0]
+
+*WARNING*: for any code relying on previously undocumented behavior of
+`state-flow.assertions.matcher-combinators/match?`, this release includes a
+breaking change.
+
+``` clojure
+;; if you were doing this before
+[actual (match? <expected> <step-that-produces-actual>)]
```
rafaelzlisboa, line 11: reading this line it wasn't clear that this was for the binding syntax - maybe wrap the example with a `flow` to make it clearer?

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -51,11 +61,21 @@
...
+                             :probe/sleep-time   ~(:sleep-time params*)
+                             :probe/times-to-try ~(:times-to-try params*))]]
+       ;; TODO: (dchelimsky, 2020-02-11) we plan to decouple
+       ;; assertions from reporting in a future release. Remove this
+       ;; next line when that happens.
        (state/wrap-fn #(~'clojure.test/testing flow-desc# (~'clojure.test/is (~'match? ~expected actual#))))
-       (state/return actual#)))))
+       (state/return report#)))))
+
+(defn report->actual
```
rafaelzlisboa, line 78: i think we don't need this fn! on the readme's example, wouldn't that work:?
```clj
[actual (fmap :match/actual (match? <expected> <step-that-produces-actual>)]
```

dchelimsky: @rafaelzlisboa yes, that would work, but that would mean committing to that key as an API, which we're not ready to do. This way, if we change anything about that map, we can change the internals of this function as well. Make sense?

CHANGELOG.md
```diff
@@ -1,7 +1,21 @@
...
-## [DEV]
-
+## [3.0.0]
+
+*WARNING*: for any code relying on previously undocumented behavior of
+`state-flow.assertions.matcher-combinators/match?`, this release includes a
+breaking change.
+
+``` clojure
+;; if you were doing this before
```
dchelimsky, line 10: How about changing this line instead, @rafaelzlisboa?
```suggestion
;; if you were doing this before in a binding
```

rafaelzlisboa: oh yeah! that works too :D

CHANGELOG.md
```diff
@@ -1,7 +1,21 @@
...
+[actual (match? <expected> <step-that-produces-actual>)]
+;; you can do this, now
+[actual (fmap report->actual (match? <expected> <step-that-produces-actual>)]
```
sovelten, line 13:
```clojure
[actual (clojure.cats.core/fmap report->actual (match? <expected> <step-that-produces-actual>))]
```
^-- missing a parenthesis at the end as well :grimacing: 

also add this other option?
```clojure
[report (match? <expected> <step-that-produces-actual>)
 :let [actual (report->actual report)]]
```

src/state_flow/assertions/matcher_combinators.clj
rafaelzlisboa: I see! I think whoever will be interested in that behaviour won't

dchelimsky: The doc for `match?` explicitly says "Returns a map (in the left value) with information about the success or failure of the match, the details of which are used internally by state-flow and subject to change." We could go on to say "if you want the actual value, use report->actual", but where does that stop?

rafaelzlisboa: But anyway, at least we're doing our part. 🚀

PR #115, opened 2020-04-27
return a vector instead of a pair for consistency

dchelimsky:


dchelimsky:
> wow, good catch!

I happened to notice it while reviewing #114

PR #116, opened 2020-04-27
Throw when times-to-try > 1 and actual is a value

dchelimsky:
Users have reported failures when probing where it turned out that `match?` was given a value instead of a step to produce the value, e.g.

```clojure
[res (do-a-step)]
(match? 37 res {:times-to-try 2})
```

This PR adds a warning when :times-to-try is > 1 and `actual` is a value instead of a step.

### DIFF
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -4,7 +4,8 @@
             [matcher-combinators.test] ;; to register clojure.test assert-expr for `match?`
             [state-flow.core :as core]
             [state-flow.probe :as probe]
-            [state-flow.state :as state]))
+            [state-flow.state :as state]
+            [taoensso.timbre :as log]))
 
 (defn ^:private match-probe
   "Internal use only.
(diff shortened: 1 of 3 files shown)


dchelimsky:
I made a choice here to warn when running the flow. Other options include:

* warn when compiling the flow
* throw (not warn) when compiling the flow
* throw (not warn) when running the flow

Anybody prefer any of those options? If so, why?

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -48,32 +49,38 @@
...
-       ;; TODO: (dchelimsky, 2020-02-11) we plan to decouple
-       ;; assertions from reporting in a future release. Remove this
-       ;; next line when that happens.
-       (state/wrap-fn #(~'clojure.test/testing flow-desc# (~'clojure.test/is (~'match? ~expected actual#))))
-       (state/return report#)))))
+     `(do
+        (when (and (> (:times-to-try ~params*) 1)
+                   (not (state/state? ~actual)))
+          (log/warn (select-keys ~caller-meta [:line])
+                    ":times-to-try > 1 has no meaningful effect when 'actual' is a value."))
```
gabrielgiussi, line 68: Maybe it is obvious for those who are used to work with `state-flow` but should we add, "'actual' should be a flow" ?

sovelten:
I don't see why someone would want to retry something that is always the same, so I think an exception makes sense here

dchelimsky:
> I don't see why someone would want to retry something that is always the same, so I think an exception makes sense here (on compiling flow)

I agree. My concern is that this is a breaking change for tests that are unnecessarily using `:times-to-try`. WDYT?

src/state_flow/assertions/matcher_combinators.clj
dchelimsky: How about "actual should be a step or a flow when :times-to-try > 1"?

sovelten:
> > I don't see why someone would want to retry something that is always the same, so I think an exception makes sense here (on compiling flow)
> 
> I agree. My concern is that this is a breaking change for tests that are unnecessarily using `:times-to-try`. WDYT?

Yeah. All the tests that explicitly use times-to-try really want it to be retried I think. I don't expect much (or at all) people using `:times-to-try` without need, but who knows. I don't see it as a breaking change.

fernando-nubank:
I agree with  @sovelten, an exception seems a better approach here.

dchelimsky:
Macroexpansion-time exception it is. Please re-review @sovelten and @fernando-nubank

test/state_flow/assertions/matcher_combinators_test.clj
```diff
@@ -96,7 +96,16 @@
       (testing "reports match-results to clojure.test"
         (is (match? {:matcher-combinators.result/type  :mismatch
                      :matcher-combinators.result/value {:expected 2 :actual 0}}
-                    (-> report-data :actual :match-result)))))))
+                    (-> report-data :actual :match-result))))))
+
+  (testing "with times-to-try > 1 and a value instead of a step"
+    (testing "throws"
+      (is (re-find #"actual must be a step or a flow when :times-to-try > 1"
```
dchelimsky, line 103: NOTE: custom exception catching because the top-level exception is CompilationError because we're throwing at macroexpansion time.

dchelimsky:
This turns out to be a breaking change for anybody upgrading from state-flow before 2.2.4 and using the old version of `match?`. The problem is that the default `times-to-try` on the old `match?` is `5`. It is then exacerbated by producing a confusing error about eval'ing locals.

Options:

1. don't throw (revert this)?
2. ¯\_(ツ)_/¯

/cc @nubank/eng-prod-test @sovelten

PR #117, opened 2020-04-29
move public API functions (including some new ones) to state-flow.core

dchelimsky:
### Why

This began with wanting to add `fmap` to state-flow's api and deciding where it belonged. In the resulting discussion we acknowledged that `state/return` is confusing because it is not a state monad-specific feature and does not interact with state at all, so we analyzed all of the different names and settled on new names we think are more intention revealing.

We also decided that the public API should all be in one namespace and landed on `state-flow.core`.

We moved implementation to a new `impl` namespace, which the deprecated functions in `state-flow.state` and the new functions in `state-flow.core` refer to. We also moved the monad context impl to new namespace to avoid circular deps.

### What

Move / rename some functions and add some new ones. All old fns remain in place, but deprecated.

* add `state-flow.core/fmap`
* mv `state-flow.state/state?` to `state-flow.core/flow?`
* mv `state-flow.state/get` and `state-flow.state/gets` to `state-flow.core/get-state`
* mv `state-flow.state/modify` to `state-flow.core/swap-state`
* mv `state-flow.state/put` to `state-flow.core/reset-state`
* mv `state-flow.state/return` to `state-flow.core/return`
* deprecate `state-flow.state/run`
* deprecate `state-flow.state/eval`
* deprecate `state-flow.state/exec`

# TODO
- [x] Add rationale (Why, above) to this PR description
- [x] update walkthrough.repl
- [x] update README
- [x] evaluate what to do with `state/run`
- [x] evaluate what to do with `state/exec`
- [x] evaluate what to do with `state/eval`
- [x] evaluate what to do with `state/state?` (maybe introduce `state-flow/flow?`)


dchelimsky:
Alternatively, we could just information about `cats.core/fmap` to the README.

dchelimsky:
See https://github.com/nubank/state-flow/pull/73/files#r375948956

dchelimsky:
@nubank/eng-prod-test - we already describe how to do a `repeat` fn using `cats.core/sequence` in the [README](https://github.com/nubank/state-flow#funcoolcats). I think we should just enhance that to include `fmap` at this point. WDYT?

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -6,6 +6,15 @@
             [state-flow.probe :as probe]
             [state-flow.state :as state]))
 
+(defn ^:private ensure-step
```
fernando-nubank, line 9: We could rename this to `ensure-flow` and refactor it to stop using deprecated functions

dchelimsky: Done

sovelten:
I like the way rewrite-clj library did here: https://github.com/xsc/rewrite-clj/blob/master/src/rewrite_clj/zip.clj

They imported the functions from other namespaces to make it easier for people to use, but kept each thing in their own namespace to make it easier to develop and maintain.

I think the primitives are in the right place, in the `state` namespace, that is why their names are `modify` and not `modify-state`. I wouldn't change the names, specially because they are based on the haskell names.

My suggestion would be to keep the names and keep the locations, import them to core.clj or other namespace we direct the user to. WDYT?

dchelimsky:
> I like the way rewrite-clj library did here: https://github.com/xsc/rewrite-clj/blob/master/src/rewrite_clj/zip.clj
> 
> They imported the functions from other namespaces to make it easier for people to use, but kept each thing in their own namespace to make it easier to develop and maintain.

Totally fine by me.

> 
> I think the primitives are in the right place, in the `state` namespace, that is why their names are `modify` and not `modify-state`. I wouldn't change the names, specially because they are based on the haskell names.

We started down this path because `return` isn't about state at all, and doesn't belong in `state` (resulting typically in `state/return`). One option would be to move only that one, but we do want them all to be exposed through the same namespace.

re: the Haskell names, I know I argued for `modify` instead of `swap` but that was based on being consistent (i.e. if we're using the Haskell names, then use all of them). The majority of the devs who encounter state-flow will not know anything about Haskell and there is value in using names that are intention revealing. IMO, the names we chose: `get-state`, `swap-state`, `reset-state`, and `return` are more intention revealing as a group than `get`, `gets`, `modify`, `put`, and `return`.

Does this rationale make sense to you @sovelten ?

> My suggestion would be to keep the names and keep the locations, import them to core.clj or other namespace we direct the user to. WDYT?

I'm OK with having the implementations in diff namespaces, all exposed through a central API namespace (maybe even `state-flow.api` instead of `state-flow.core`).

sovelten:
> I'm OK with having the implementations in diff namespaces, all exposed through a central API namespace (maybe even `state-flow.api` instead of `state-flow.core`).

I don't feel strongly for the names as long as they are internally consistent. I do feel an obsession for avoiding repeating stuff, I wonder if it is to awkward to use state-flow.state/swap internally and redefine to `state-flow.api/swap-state` in the API

I would think about these changes carefully because we already went through one such change, and it is not so easy to know what really is more intuitive.

dchelimsky:
>I don't feel strongly for the names as long as they are internally consistent. I do feel an obsession for avoiding repeating stuff, I wonder if it is too awkward to use state-flow.state/swap internally and redefine to state-flow.api/swap-state in the API

Agreed on not repeating. I'm not 100% clear what you're saying here, but we deprecated the fns in `state-flow.state`, so having our new API fns point at them would be awkward. That's why we have the deprecated fns use the new ones. Does that make sense?

> I would think about these changes carefully because we already went through one such change, and it is not so easy to know what really is more intuitive.

Agreed! That's why we're doing this whole exercise.

philomates:
I don't have many opinions that follow logical arguments, but I'll drop by to offer some intuitive thoughts:

 - I also appreciate names like `foo/get` in the place of `bar/get-foo`
 - If we put everything in one namespace, will we be doing a `[state-flow.core :refer :all]`? If so, do we care about clobbering core clojure functions like `swap`?
 - Perhaps a dumb idea, but what if `core` was renamed `flow`, and then there was a `state-flow.flow` and `state-flow.state` dichotomy? `core` doesn't really tell you what is in `core` at the moment; though I guess it does tell you were to start. Is the argument of putting everything in one namespace to improve user ergonomics at require time?
 - Given how Nubank uses `state-flow`, you usually don't need to mess with the state. Thus it is tempting to suggest leaving such state manipulation functions in a dedicated namespace. Sort of as a way to signify that this is a new domain that requires new knowledge.
 - Both `swap`/`reset` and `modify`/`put` have pros and cons. `swap`/`reset` are more descriptive, but make one lean on clojure experience, which might prevent one from fully embracing the monadic context (just a guess). Whereas `modify`/`put` have Haskell intuition (which means not much here) and is also distinct from vanilla clojure, though maybe that just adds to mental overhead of remembering both argument types + function.
 - I find `return` to be an unfortunate name. Given that here it is "lifting" a pure computation (as in a value) into the flow context, `lift` sounds compelling to me, but of course is conflated with Haskell's monadic `lift`

dchelimsky:
> I don't have many opinions that follow logical arguments, but I'll drop by to offer some intuitive thoughts:
> 
>     * I also appreciate names like `foo/get` in the place of `bar/get-foo`

I agree in general. In this case, we went from `return` doesn't make sense in `state` to "let's move everything exposed from one namespace. Once we got there, `get` doesn't make as much sense.

>     * If we put everything in one namespace, will we be doing a `[state-flow.core :refer :all]`? If so, do we care about clobbering core clojure functions like `swap`?

I see `[xxx :refer :all]` as an anti-pattern. 

> 
>     * Perhaps a dumb idea, but what if `core` was renamed `flow`, and then there was a `state-flow.flow` and `state-flow.state` dichotomy? `core` doesn't really tell you what is in `core` at the moment; though I guess it does tell you were to start. Is the argument of putting everything in one namespace to improve user ergonomics at require time?

I like this dumb idea!

>     * Given how Nubank uses `state-flow`, you usually don't need to mess with the state. Thus it is tempting to suggest leaving such state manipulation functions in a dedicated namespace. Sort of as a way to signify that this is a new domain that requires new knowledge.

Makes sense.

>     * Both `swap`/`reset` and `modify`/`put` have pros and cons. `swap`/`reset` are more descriptive, but make one lean on clojure experience, which might prevent one from fully embracing the monadic context (just a guess). Whereas `modify`/`put` have Haskell intuition (which means not much here) and is also distinct from vanilla clojure, though maybe that just adds to mental overhead of remembering both argument types + function.

We discussed this at length today in patota-test and landed on this: monad is an implementation detail of state-flow, and while we don't want to hide that fact, we also don't want to make it important to understand for people to use state-flow.

Part of the problem is the conversation has been constrained by existing names in both languages, which are naturally overloaded. What if we chose names are from neither Clojure nor Haskell, e.g.

```clojure
(state/push f & args)
(state/pull f & args)
```

>     * I find `return` to be an unfortunate name. Given that here it is "lifting" a pure computation (as in a value) into the flow context, `lift` sounds compelling to me, but of course is conflated with Haskell's monadic `lift`

In state-flow, `return` is only useful for bindings, which you can do this way:

```clojure
[:let [sym expr]]
```

I think we can just deprecate it.

On the topic of bindings, some editors struggle with the vector syntax. WDYT about introducing a `bind` (or similar name) fn:

```clojure
(flow "do stuff"
  (bind [sym expr-or-flow])
  (do-something-with sym))
```

I know that @sovelten has expressed that the fact that `match?` can work with a value or a flow is undesirable magic, but this would solve a lot of problems (smaller syntax that is Clojure syntax).

dchelimsky:
To clarify my last comments (in response to @philomates). I'm proposing that we have three functions:

* create a step that returns the application of a function to state
* create a step that stores the application of a function to state
* bind a value (or the return of a flow/step) to a symbol

Also:
* their names should avoid conflicting with overloaded names in Clojure and Haskell
* syntax should be Clojure-editor friendly

rafaelzlisboa:
> Perhaps a dumb idea, but what if core was renamed flow, and then there was a state-flow.flow and state-flow.state dichotomy? core doesn't really tell you what is in core at the moment; though I guess it does tell you were to start. Is the argument of putting everything in one namespace to improve user ergonomics at require time?

I think user ergonomics and ease of exploring is super important. state-flow can be simple enough so that `core` can be `state-flow` in people's minds and they don't need to poke into internal namespaces.
Most of the times folks writing integration tests are already juggling a bunch of helpers namespaces, I think we should do whatever's possible to simplify :)

sovelten:
> In state-flow, `return` is only useful for bindings, which you can do this way:
> 
> ```clojure
> [:let [sym expr]]
> ```
> 
> I think we can just deprecate it.

That's not true, return is used to signify that a flow returns some specified value, specially when you want to return something different from the last flow. return is used on many state-flow-helpers functions. And as you can see, I used the word "return" to explain what return does, so I do think there is a good intuition to the name. I know that it could be misleading if you think it would be the same as `return` reserved keyword from many languages, but I think it is much more intention revealing than `lift`. I would keep it as is.

> 
> On the topic of bindings, some editors struggle with the vector syntax. WDYT about introducing a `bind` (or similar name) fn:
> 
> ```clojure
> (flow "do stuff"
>   (bind [sym expr-or-flow])
>   (do-something-with sym))
> ```
> 
> I know that @sovelten has expressed that the fact that `match?` can work with a value or a flow is undesirable magic, but this would solve a lot of problems (smaller syntax that is Clojure syntax).

Yeah, I'm still against this. The editors should not be a reason for limiting syntax expressiveness, it sounds as a poor reason to me. Using `:let` is the same syntax used in `for`, same as `mlet`. In Haskell there is a clear distinction between a monadic bind `a <- b` and a pure bind `let a = b`. Just because you are a dynamic language, doesn't mean that you should accept anything and modify behaviour accordingly, it generates more confusion than it solves. Also, bind is an overloaded term, because bind is the main function of the monad protocol.

dchelimsky:
> That's not true, return is used to signify that a flow returns some specified value, specially when you want to return something different from the last flow.

We should call out use cases like this in the README.

dchelimsky:
I opened a new PR for just fmap, return, and invoke: #118. Closing this one.

PR #118, opened 2020-05-11
add state-flow.api namespace

dchelimsky:
This is take 2 of #117, which originally started as "let's add fmap somewhere"

## state-flow.api

Introduce a new `state-flow.api` namespace which imports the following vars:

``` clojure


;; from state-flow.core
flow
run
run*
log-and-throw-error!
ignore-error

;; from state-flow.state
get-state  ;; state/gets
swap-state ;; state/modify
invoke     ;; formerly wrap-fn
return
fmap       ;; new addition

;; from state-flow.cljtest
defflow

;; from state-flow.assertions.matcher-combinators/
match?
```

## Deprecates

* state/wrap-fn (replaced by state/invoke)

## Also

* improves/clarifies some docstrings

## Rationale:

We wanted to add `fmap` and decided it made more sense in `core` than in `state`. While reviewing this, we realized that `state/return` and `state/wrap-fn` have little to do with state, so they probably belonged in `core` as well, and we found that the name `wrap-fn` was confusing, hence `invoke`.

The same conversation also revealed that there is some pain in having to require 4 namespaces in order to use state-flow in its most common use at Nubank: with clojure.test and matcher-combinators. That's what led us to the `api` namespace.

## TODO

- [x] consider recommending aliasing `[state-flow.api :as flow]`
  - that would put nice things like `flow/invoke` and `flow/return` in flows.
- [x] update README
- [x] update walkthrough

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [4.0.0]
+
+* Add `state-flow.api` namespace [#118](https://github.com/nubank/state-flow/pull/118)
+  * New namespace which has everything you need™
+  * New `fmap` fn
+
 ## [3.1.0]
 
 * Add state-flow.core/runner to access runner within flows [#119](https://github.com/nubank/state-flow/pull/119)
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -113,33 +113,6 @@
   [s]
   (-> s meta :top-level-description))
 
-(defn return
-  "Creates a flow that returns v. Use this as the last
-  step in a flow that you want to reuse in other flows, in
-  order to clarify the return value, e.g.
-
-    (def increment-count
-      (flow \"increments :count and returns it\"
-        (state/modify update :count inc)
-        [new-count (state/gets :count)]
-        (state-flow/return new-count)))"
-  [v]
-  (m/return state/short-circuiting-context v))
-
-(def
-  ^{:doc "Creates a flow that returns the application of f to the return of flow"
-    :arglists '([f flow])}
-  fmap
-  m/fmap)
-
-(defn invoke
-  "Creates a flow that invokes a function of no arguments and returns the
-  result. Used to invoke side effects e.g.
-
-     (state-flow.core/invoke #(Thread/sleep 1000))"
-  [my-fn]
-  (state/error-catching-state (fn [s] [(my-fn) s])))
-
 (defn ^:deprecated as-step-fn
   "DEPRECATED with no replacement."
   [flow]
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -80,7 +80,15 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Public API
 
-(defn flow* [{:keys [description caller-meta]} & flows]
+(defn flow*
+  "For use in macros that create flows. Not private (appropriate for
+  helper libraries, for example), but not intended for use directly in
+  flows.
+
+  Creates a flow which is a composite of flows. The calling macro should
+  provide (meta &form) as `:caller-meta` in order to support accurate line
+  number reporting."
+  [{:keys [description caller-meta]} & flows]
   (when-not (string-expr? description)
     (throw (IllegalArgumentException. "The first argument to flow must be a description string")))
   (let [flow-meta caller-meta
@@ -92,7 +100,7 @@
       (m/return ret#))))
 
 (defmacro flow
-  "Defines a flow"
+  "Creates a flow which is a composite of flows."
   {:style/indent :defn}
   [description & flows]
   (apply flow* {:description description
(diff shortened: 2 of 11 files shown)


sovelten:
Does it make sense to move this to state-flow.api now? That will probably save another deprecation.

Still looks inconsistent to me, `return` is part of the primitives.
Of course you could simulate `(return v)` with `(state/gets (constantly v))`, but that's not the point.

I would keep `return` and `invoke` exactly where they are and create a `state-flow.api` where we can export `return`, `invoke`, `modify` (or whatever name)

From the library development perspective it makes sense to keep state primitives in the state-flow.state namespace because they are the building blocks on top of which we build state-flow itself. From the user perspective it makes sense to have all the user level api in a single namespace. Can we have our cake and eat it too?

dchelimsky:
@sovelten We're definitely interested in all forms of possessing and consuming :cake:!

There's a disconnect between your view of `return` and the one a few of us on test-pack have developed. You're saying it's part of the primitives, but we see it (and invoke) as fundamentally different from e.g. `get` and `modify` as it does not actually interact with state in any way:


That's because `get` and `modify` get and modify state, whereas `return` and `invoke` have nothing to do with state. If we leave them all in the same namespace, then they don't make as much sense:

``` clojure
flow/get    ;; get what?
flow/modify ;; modify what?
flow/return
flow/invoke

api/get    ;; get what?
api/modify ;; modify what?
api/return
api/invoke
```

We think this organization is more intention revealing:
nse (to us):

```clojure
state/get    ;; oh! We're getting state!
state/modify ;; oh! We're getting state!
flow/return
flow/invoke
```

Is that clear?

sovelten:
> We think this organization is more intention revealing:
> nse (to us):
> 
> ```clojure
> state/get    ;; oh! We're getting state!
> state/modify ;; oh! We're getting state!
> flow/return
> flow/invoke
> ```
> 
> Is that clear?

That's quite clear, I'm just not sure I agree.

(state/return v) -> (State. (fn [s] [v s]))
(state/invoke f) -> (State. (fn [s] [(f) s]))

Even though they don't change the state, they deal directly with the state transition machinery and data structure defined in the state ns, one we don't want to expose the rest of the library to.

Let me say first of all that `state monad` is a terrible name for this monad (naming is still the most difficult thing in computer science (and monad is a terrible name for this kind of abstraction (oh well...))), because the data structure is not a state, it is a state transition function (that may or may not change the state) and that may be the reason for the apparent disconnection of `return` and `invoke`. It's not about changing state, it is about composing state aware functions.

That's why I propose to keep `return` and `invoke` where they are and create an api in which we can do:

```clojure
(flow/invoke ...)
(flow/get-state ...) ; names may vary
(flow/swap-state ...)
(flow/return ...)
```

I believe this is having the cake and eating it.

dchelimsky:
> That's why I propose to keep `return` and `invoke` where they are and create an api in which we can do:
> 
> ```clojure
> (flow/invoke ...)
> (flow/get-state ...) ; names may vary
> (flow/swap-state ...)
> (flow/return ...)
> ```
> 
> I believe this is having the cake and eating it.

That works for me, personally, but @philomates expressed a distaste for fns named `get-xxx` vs `xxx/get`. @philomates can you be convinced this is an astoundingly beautiful approach?

dchelimsky:
> Let me say first of all that `state monad` is a terrible name for this monad (naming is still the most difficult thing in computer science (and monad is a terrible name for this kind of abstraction (oh well...))), because the data structure is not a state, it is a state transition function (that may or may not change the state) and that may be the reason for the apparent disconnection of `return` and `invoke`. It's not about changing state, it is about composing state aware functions.

I don't see this fully (though I'm trying). What is "state aware" about `return` and `invoke`? Are you saying they're state-aware, but just choose to ignore state (making them state-ignorant and state-aware at the same time!)?

sovelten:
> > Let me say first of all that `state monad` is a terrible name for this monad (naming is still the most difficult thing in computer science (and monad is a terrible name for this kind of abstraction (oh well...))), because the data structure is not a state, it is a state transition function (that may or may not change the state) and that may be the reason for the apparent disconnection of `return` and `invoke`. It's not about changing state, it is about composing state aware functions.
> 
> I don't see this fully (though I'm trying). What is "state aware" about `return` and `invoke`? Are you saying they're state-aware, but just choose to ignore state (making them state-ignorant and state-aware at the same time!)?

Yeah... I guess? :sweat_smile: 
Let's call this thing blorb. All that `get`, `put`, `return`, `invoke` do is create raw blorb. We don't want to deal with blorb directly because blorb is ugly and also it would be an abstraction leak, we want blorb contained, we don't want people ever to know that they can create blorb with `(State. (fn [s] ...)`, not even state-flow.core, because if you don't use the api you might end up with a corrupted/useless blorb. So instead we use those functions as a blorb api to creating and composing blorb.

Now, the blorb api just happens to also be useful for composing flows (because flows deep down are just fancy blorbs). So we can expose part or all of the blorb api to the user. Hope that all makes sense now.

dchelimsky:
> Now, the blorb api just happens to also be useful for composing flows (because flows deep down are just fancy blorbs). So we can expose part or all of the blorb api to the user. Hope that all makes sense now.

Not a bit, but thanks for trying ;)

dchelimsky:
> That's why I propose to keep `return` and `invoke` where they are and create an api in which we can do:
> 
> ```clojure
> (flow/invoke ...)
> (flow/get-state ...) ; names may vary
> (flow/swap-state ...)
> (flow/return ...)
> ```
> 
> I believe this is having the cake and eating it.

@sovelten as ordens!

I think this is great now. Still need to update the README and walkthrough, but we're getting close.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Changelog
 
-## [DEV]
+## [4.0.0]
 
+* Add `state-flow.api` namespace [#118](https://github.com/nubank/state-flow/pull/118)
+  * New namespace which has everything you need™
```
dchelimsky, line 6:
```suggestion
  * New namespace which has EverythingYouNeed™
```

src/state_flow/core.clj
```diff
@@ -92,7 +100,7 @@
       (m/return ret#))))
 
 (defmacro flow
-  "Defines a flow"
+  "Creates a flow which is a composite of flows."
```
philomates, line 103: :exploding_head: Xzibit approves

PR #119, opened 2020-05-14
Add state-flow.state/runner to access the runner within flows

dchelimsky:


test/state_flow/state_test.clj
```diff
@@ -2,6 +2,7 @@
   (:require [cats.core :as m]
             [cats.monad.exception :as e]
             [clojure.test :as t :refer [deftest is testing]]
+            [state-flow.core :as state-flowo]
```
rafaelzlisboa, line 5: typo

gabrielgiussi:
Wich is the use case of this? When I could want to access the runner from a flow?

dchelimsky:
> Which is the use case of this? When I could want to access the runner from a flow?

It's for helpers that need to run a flow within a flow in order to wrap it in some other function (e.g. `with-redefs`.

src/state_flow/state.clj
```diff
@@ -113,3 +113,9 @@
   (if (state? value)
     value
     (return value)))
+
+(defn runner
```
sovelten, line 117: Now this is part of core, no? runner is defined in core and meta is also something related to core.

test/state_flow/state_test.clj
```diff
@@ -2,6 +2,7 @@
   (:require [cats.core :as m]
             [cats.monad.exception :as e]
             [clojure.test :as t :refer [deftest is testing]]
+            [state-flow.core :as state-flow]
```
sovelten, line 5: You shouldn't declare core on state_test.clj namespace

PR #123, opened 2020-05-18
with-redefs + wrap-with macros

caioaao:


src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,19 @@
...
+(defmacro with-redefs
+  [bindings flow]
+  (assert *enable-with-redefs-macro*
+          "`with-redefs` usage is not recommended. If you know what you're doing and really want to continue, set `*enable-with-redefs-macro*` to true")
+  `(m/do-let
+    [world#  (state/get)
+     runner# (state-flow/runner)
+     :let [[ret# state#] (clojure.core/with-redefs ~bindings
+                           (runner# ~flow world#))]]
+    (state/put state#)
```
dchelimsky, line 18:
```suggestion
    (state-flow.api/swap-state (constantly state#))
```

src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,19 @@
...
+    (state/return ret#)))
```
dchelimsky, line 19:
```suggestion
    (state-flow.api/return ret#)))
```

dchelimsky:
I think if we're going to add this to state-flow proper (even in labs), I think we should add the `wrap-in` generalization. The you could use e.g.

```clojure
(wrap-in (fn [wrapped-flow] (with-redefs [<bindings>] wrapped-flow))
         flow-to-wrap))
```

src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,19 @@
...
+  (:refer-clojure :exclude [with-redefs])
+  (:require [cats.core :as m]
+            [state-flow.state :as state]
+            [state-flow.core :as state-flow]))
+
+(def ^:dynamic *enable-with-redefs-macro* nil)
+
+(defmacro with-redefs
+  [bindings flow]
+  (assert *enable-with-redefs-macro*
```
sovelten, line 11: Not sure about the usage of this variable. Just a warning in the comments is fine. The flag looks like a handicap to me.

src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,19 @@
+(ns state-flow.labs.state
```
sovelten, line 1: This would probably go better in another namespace. Maybe `labs.helpers`

dchelimsky:
> I think if we're going to add this to state-flow proper (even in labs), I think we should add the wrap-in generalization.

Background: https://github.com/nubank/state-flow/pull/33

src/state_flow/labs/state.clj
caioaao: I thought it'd go along with the other state fns, like `wrap-fn`

caioaao:
> I think if we're going to add this to state-flow proper (even in labs), I think we should add the wrap-in generalization. 

@dchelimsky from what I saw, #33 was closed because of the potential of misuse. If we limit this implementation to specific use cases we see the need for it, the misuse might be reduced, no? We can keep `wrap-with` but as a private function

caioaao:
We can also leave it public for now in labs and check if this is going to be an issue

src/state_flow/labs/state.clj
dchelimsky: I think it's fine here, but we should have a docstring in this namespace that indicates that it is experimental and subject to change.

src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,25 @@
...
+  (:refer-clojure :exclude [with-redefs])
+  (:require [cats.core :as m]
+            state-flow.api
+            [state-flow.core :as state-flow]
+            [state-flow.state :as state]))
+
+(defmacro wrap-with
+  "WARNING: `wrap-with` usage is not recommended. Use only if you know what you're
```
dchelimsky, line 9: This warning is fine, but we should also say what it does :-D

src/state_flow/labs/state.clj
```diff
@@ -0,0 +1,25 @@
...
+  [wrapper-fn flow]
+  `(m/do-let
+    [world#  (state/get)
+     runner# (state-flow/runner)
+     :let [[ret# state#] (~wrapper-fn (fn [] (runner# ~flow world#)))]]
+    (state-flow.api/swap-state (constantly state#))
+    (state/return ret#)))
+
+(defmacro with-redefs
+  "WARNING: `with-redefs` usage is not recommended. Use only if you know what you're
```
dchelimsky, line 20: Same comment - docstring should also say what this does, and warn that `with-redefs` is not threadsafe.

test/state_flow/labs/state_test.clj
```diff
@@ -0,0 +1,41 @@
...
+(defn wrap-with-redefs [f]
+  (with-redefs [put2 put3]
+    (f)))
+
+(defn called-it-callback [f]
+  (print "called it")
+  (f))
+
+(deftest wrap-with-test
+  (testing "verify wrapper is called"
```
dchelimsky, line 20:
```suggestion
  (testing "wrapper is called"
```

test/state_flow/labs/state_test.clj
```diff
@@ -0,0 +1,41 @@
...
+    (is (= "called it"
+           (with-out-str
+             (-> (labs.state/wrap-with
+                  (fn [f] (print "called it") (f))
+                  (state/modify put2))
+                 (state-flow/run {}))))))
+  (testing "verify flow runs successfully"
```
dchelimsky, line 27:
```suggestion
  (testing "flow runs successfully"
```

test/state_flow/labs/state_test.clj
```diff
@@ -0,0 +1,41 @@
...
+    (is (match? [{} {:value 2}]
+                (-> (labs.state/wrap-with
+                  (fn [f] (f))
+                  (state/modify put2))
+                 (state-flow/run {}))))))
+
+(deftest with-redefs-test
+  (testing ""
```
dchelimsky, line 35: Either fill in the docstring or bag `testing`

src/state_flow/labs/state.clj
dchelimsky: I think we should say that this here to support helper macros like `with-redefs`, and reference that as an implementation example.

dchelimsky:
Thanks for all the updates @caioaao. I'm waiting for some feedback from other folks within Nubank.

fernando-nubank:
> We can also leave it public for now in labs and check if this is going to be an issue

I have more issues with `with-redefs` than with `wrap-fn` but, as both of them are in `labs` I think they can be public.

dchelimsky:
> LGTM now, but I want at least one other approval before merging.

Oh! There is one. I'm going to add a commit to update the changelog first. Then I'll merge it.

dchelimsky:
> > LGTM now, but I want at least one other approval before merging.
> 
> Oh! There is one. I'm going to add a commit to update the changelog first. Then I'll merge it.

I can't do that on  your branch @caioaao so I'll do it in a follow up PR before releasing.

PR #125, opened 2020-05-18
move match? exception to flow execution time

dchelimsky:
Before this PR `match?` would raise an exception at macroexpansion time if it was called in a syntactically correct by semantically incorrect way. With this PR, the update happens at flow execution time instead.

Additionally, in order to reduce upgrade friction, that exception is not thrown at all when calling the deprecated `state-flow.cljtest/match?` fn.

### DIFF
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -10,7 +10,8 @@
   (let [params* (merge {:times-to-try probe/default-times-to-try
                         :sleep-time   probe/default-sleep-time
                         :caller-meta  (meta &form)
-                        :description  match-desc}
+                        :description  match-desc
+                        :called-from-deprecated-match? true}
                        params)]
     `(~'state-flow.assertions.matcher-combinators/match? ~expected ~actual ~params*)))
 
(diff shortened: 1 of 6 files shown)


test/state_flow/assertions/matcher_combinators_test.clj
```diff
@@ -101,11 +101,12 @@
   (testing "with times-to-try > 1 and a value instead of a step"
     (testing "throws"
       (is (re-find #"actual must be a step or a flow when :times-to-try > 1"
-                   (.. (try
-                         (eval `(mc/match? 3 (+ 30 7) {:times-to-try 2}))
```
dchelimsky, line 105: `eval` here was a hint that we were doing it wrong.

src/state_flow/cljtest.clj
```diff
@@ -10,7 +10,8 @@
   (let [params* (merge {:times-to-try probe/default-times-to-try
                         :sleep-time   probe/default-sleep-time
                         :caller-meta  (meta &form)
-                        :description  match-desc}
+                        :description  match-desc
+                        :called-from-deprecated-match? true}
```
dchelimsky, line 14: This is used to tell the non-deprecated version of match? not to throw the `times-to-throw` exception so that when people upgrade from state-flow < 2.2.4, they don't start seeing exceptions right away.

dchelimsky: The rationale here is that the old default for times-to-try is `5`, so the old match? was raising exceptions in a lot of cases by default.

Issue #113, opened 2020-04-20
Improve error message when last flow form is a vector


Andre0991:
`state-flow` won't compile this:

```
(defflow foo
  (flow "consuming `installments-loan.new`"
    [:let [as-of #nu/time "2020-01-07T13:00:00Z"]]))
```

It gets this error:

```
Unexpected error (AssertionError) macroexpanding cats.core/do-let at (postman/postman/events/update_pv_for_closing_bill.clj:50:7).
Assert failed: Last argument of do-let must not be a vector
(not (vector? (last forms)))
```

It would be nice if it allowed such forms.

It makes no sense for the test, but it's useful for interactive development.

dchelimsky:
A flow must return a flow/step, and we don't want to change that semantic. We can, however, improve the error message. Leaving this open until we do.

Andre0991:
Hummm, yeah, that makes sense. Thanks.

dchelimsky:
Addressed in PR #128

PR #128, opened 2020-06-08
improve error message when last arg to flow is a vector

dchelimsky:
This PR addresses issue #113 

### Before

```clojure
(flow "description" [x (state/get-state)])
;; Unexpected error (AssertionError) macroexpanding cats.core/do-let at (....).
;; Assert failed: Last argument of do-let must not be a vector
;; (not (vector? (last forms)))
```

### After

```clojure
(flow "description" [x (state/get-state)])
;; Syntax error macroexpanding state-flow.core/flow at (...).
;; The last argument to flow must be a flow/step, not a binding vector.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [5.1.0]
+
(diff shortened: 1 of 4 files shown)


dchelimsky:
Thanks @philomates

PR #129, opened 2020-06-11
Add error handler to filter stack trace

dchelimsky:
This PR:

* breaks up `log-and-throw-error!` into `log-error` and `throw-error`
    * `log-and-throw-error!` is deprecated, not removed
* makes the default `:on-error` in `state-flow.api/run*` a composition of:
    `(comp throw-error! log-error (filter-stack-trace default-stack-trace-exclusions))`
* you can override this with any of those handlers or any comp of them e.

```clojure
(run* {:on-error ignore-error} flow)
(run* {:on-error (comp log-error (filter-strack-trace))} flow)
```

* you can also override the stack-trace-exclusions e.g.

```clojure
(run* {:on-error (comp log-error (filter-strack-trace [#"clojure.lang"]))} flow)
``` 

### Before (defaults)

```clojure
user> (flow/run* {} (flow/flow "" (flow/invoke (/ 1 0))))
20-06-15 15:23:48 MacBook-Pro.local INFO [state-flow.core:142] - Flow "" failed with exception
nrepl.middleware.interruptible-eval/evaluate/fn  interruptible_eval.clj:   91
                              clojure.core/eval                core.clj: 3214
                                            ...
                                 user/eval26881               REPL Input
                           state-flow.core/run*                core.clj:  233
                            state-flow.core/run                core.clj:  205
                           cats.monad.state/run              state.cljc:  147
       state-flow.state/error-catching-state/fn               state.clj:   20
                                            ...
                   cats.monad.exception/wrap/fn          exception.cljc:  250
               cats.monad.exception/exec-try-on          exception.cljc:  197
      cats.monad.exception/wrap/fn/func--auto--          exception.cljc:  250
                             clojure.core/apply                core.clj:  665
                                            ...
                      state-flow.state/reify/fn               state.clj:   50
                              cats.core/bind/fn               core.cljc:   96
                              user/eval26881/fn               REPL Input
                                            ...
java.lang.ArithmeticException: Divide by zero

Execution error (ArithmeticException) at user/eval26881$fn (form-init8086161948397246332.clj:225).
Divide by zero
```

### After (defaults)

``` clojure
user> (flow/run* {} (flow/flow "" (flow/invoke (/ 1 0))))
20-06-15 15:23:05 MacBook-Pro.local INFO [state-flow.core:142] - Flow "" failed with exception
                       clojure.core/eval   core.clj: 3214
                          user/eval26794  REPL Input
                    state-flow.core/run*   core.clj:  233
                     state-flow.core/run   core.clj:  205
state-flow.state/error-catching-state/fn  state.clj:   20
                      clojure.core/apply   core.clj:  665
               state-flow.state/reify/fn  state.clj:   50
                       user/eval26794/fn  REPL Input
                                     ...
java.lang.ArithmeticException: Divide by zero

Execution error (ArithmeticException) at user/eval26794$fn (form-init8086161948397246332.clj:210).
Divide by zero

```

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -219,7 +219,7 @@
 
                     `(comp throw-error!
                            log-error
-                           (filter-stack-trace))`"
+                           (filter-stack-trace default-strack-trace-exclusions))`"
   [{:keys [init cleanup runner on-error]
     :or   {init                   (constantly {})
            cleanup                identity
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -144,29 +144,32 @@
   pair)
 
 (def default-stack-trace-exclusions
-  (let [aviso-defaults [#"clojure.lang"
-                        #"sun\.reflect.*"
-                        #"java.lang.reflect"
-                        #"speclj\..*"
-                        #"clojure\.main/repl/read-eval-print.*"]]
-    (into aviso-defaults
-          [#"^clojure\.main"
-           #"^state_flow\."
-           #"^nrepl\."
-           #"^cats\.."
-           #"^io\.pedestal\..*"])))
+  [#"^nrepl\."
+   #"^cats\."
+   #"^java\.lang\.reflect"
+   #"^java\.lang\.Thread"
+   #"^clojure\.main\$repl"
+   #"^clojure\.lang"])
 
-(defn filter-stack-trace* [filter throwable]
-  (let [lines     (into [] (.getStackTrace throwable))
-        filtered  (remove
-                  (fn [line] (some #(re-find % (.getClassName line)) filter))
-                  lines)]
-    (doto throwable (.setStackTrace (into-array filtered)))))
+(defn filter-stack-trace
+  "Given a seq of exclusions (regexen) and a StackTraceElement array,
+  returns a new StackTraceElement array which excludes all elements
+  whose class names match any of the exclusions."
+  [exclusions stack-trace]
+  (let [frames (into [] stack-trace)]
+    (->> (into [(first frames)]
+               (remove
+                (fn [frame]
+                  (some #(re-find % (.getClassName frame)) exclusions))
+                (rest frames)))
+         into-array)))
 
-(defn filter-stack-trace [filter pair]
+(defn with-filtered-stack-trace [exclusions pair]
   (if-let [failure (some->> pair first :failure)]
     [(#'cats.monad.exception/->Failure
-      (filter-stack-trace* filter failure))
+      (doto failure
+        (.setStackTrace
+         (filter-stack-trace exclusions (.getStackTrace failure)))))
      (second pair)]
     pair))
 
@@ -208,7 +211,7 @@
   (let [initial-state (init)
         pair          (->> (runner flow (vary-meta initial-state assoc :runner runner))
                            clarify-illegal-arg
-                           (filter-stack-trace stack-trace-exclusions))]
+                           (with-filtered-stack-trace stack-trace-exclusions))]
     (try
       (cleanup (second pair))
       pair
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -143,6 +143,33 @@
   [pair]
   pair)
 
+(def default-stack-trace-exclusions
+  (let [aviso-defaults [#"clojure.lang"
+                        #"sun\.reflect.*"
+                        #"java.lang.reflect"
+                        #"speclj\..*"
+                        #"clojure\.main/repl/read-eval-print.*"]]
+    (into aviso-defaults
+          [#"^clojure\.main"
+           #"^state_flow\."
+           #"^nrepl\."
+           #"^cats\.."
+           #"^io\.pedestal\..*"])))
+
+(defn filter-stack-trace* [filter throwable]
+  (let [lines     (into [] (.getStackTrace throwable))
+        filtered  (remove
+                  (fn [line] (some #(re-find % (.getClassName line)) filter))
+                  lines)]
+    (doto throwable (.setStackTrace (into-array filtered)))))
+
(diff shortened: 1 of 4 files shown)


src/state_flow/core.clj
```diff
@@ -143,6 +143,33 @@
   [pair]
   pair)
 
+(def default-stack-trace-exclusions
+  (let [aviso-defaults [#"clojure.lang"
```
dchelimsky, line 147: state-flow does not use aviso, but I wanted to identify this list. Thoughts?

src/state_flow/core.clj
```diff
@@ -163,23 +190,28 @@
...
-           cleanup  identity
-           runner   run
-           on-error log-and-throw-error!}
-    :as opts}
+    `:init`                   optional, default (constantly {}), function of no arguments that returns the initial state
+    `:cleanup`                optional, default identity, function of the final state used to perform cleanup, if necessary
+    `:runner`                 optional, default `run`, function of a flow and an initial state which will execute the flow
+    `:on-error`               optional, default `log-and-throw-error!`, function of the final result pair to be invoked
+                              when the first value in the pair represents an error
+    `:stack-trace-exclusions` optional, default `default-stack-trace-exclussions`, sequence of regular expressions
```
dchelimsky, line 198: I realize this name is long, but calling it a `filter` would suggest that it's the things we keep rather than the things we remove.

src/state_flow/core.clj
rafaelzlisboa: what's the main difference between what's on this list versus what's on the vector below?

src/state_flow/core.clj
```diff
@@ -143,6 +143,33 @@
   [pair]
   pair)
 
+(def default-stack-trace-exclusions
+  (let [aviso-defaults [#"clojure.lang"
+                        #"sun\.reflect.*"
+                        #"java.lang.reflect"
+                        #"speclj\..*"
```
rafaelzlisboa, line 150: i think this one doesn't need to be a default

src/state_flow/core.clj
dchelimsky: I'm not sure I understand your question, since you can see both vectors right here. The aviso defaults are just that, the regular expressions aviso uses to excluse stack trace elements by default. The additions are ... additions.

src/state_flow/core.clj
dchelimsky: If we're going to base this on the aviso defaults, it does. We could ignore aviso here, but this is part of their list, which I presume comes from some research.

dchelimsky: Do you think we should or should not care about the aviso list?

src/state_flow/core.clj
```diff
@@ -163,23 +193,28 @@
...
-           cleanup  identity
-           runner   run
-           on-error log-and-throw-error!}
-    :as opts}
+    `:init`                   optional, default (constantly {}), function of no arguments that returns the initial state
+    `:cleanup`                optional, default identity, function of the final state used to perform cleanup, if necessary
+    `:runner`                 optional, default `run`, function of a flow and an initial state which will execute the flow
+    `:on-error`               optional, default `log-and-throw-error!`, function of the final result pair to be invoked
+                              when the first value in the pair represents an error
+    `:stack-trace-exclusions` optional, default `default-stack-trace-exclussions`, sequence of regular expressions
```
sovelten, line 201: I wonder if this is just a part of the error handler instead of a new option:
`:on-error (comp log-and-throw-error! filter-stack-trace)`

dchelimsky: It could be, but we need to be able to specify the filters, so maybe:

```clojure
:on-error (comp log-and-throw-error! (filter-stack-trace [<exclusions>]))
```

The trick here is that `log-and-throw-error!` and `filter-stack-trace` have different contracts, so you'd have to get the order right. The implementation in this PR avoids that. WDYT?

sovelten: Well, you have to get the order right because the last one throws, and it's clear that it throws by the bang and the name. I like the compositional approach better, but I'm fine either way.

dchelimsky: Thinking of composition, `log-and-throw-error!` has an "and" in its name! WDYT about:

```clojure
:on-error (comp throw-error! log-error (filter-stack-trace [<exclusions>]))
```

dchelimsky:
I converted this back to a Draft to avoid an optimistic merge until we've resolved the discussion.

src/state_flow/core.clj
sovelten: Yes, please! That's what I do https://github.com/nubank/purgatory/blob/master/src/purgatory/diplomat/consumer.clj#L52

dchelimsky: OK. Gonna have to wait for Monday, though ;)

dchelimsky: @sovelten please re-re-review.

src/state_flow/core.clj
```diff
@@ -163,23 +211,30 @@
...
-               when the first value in the pair represents an error"
+    `:init`       optional, default (constantly {}), function of no arguments that returns the initial state
+    `:cleanup`    optional, default identity, function of the final state used to perform cleanup, if necessary
+    `:runner`     optional, default `run`, function of a flow and an initial state which will execute the flow
+    `:on-error`   optional, function of the final result pair to be invoked when the first value in the pair
+                  represents an error, default:
+
+                    `(comp throw-error!
+                           log-error
+                           (filter-stack-trace))`"
```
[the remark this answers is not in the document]
dchelimsky: `filter-stack-trace` has a 1-arg arity that uses `default-stack-trace-exclusions`, so that code is legit. Now that you bring it up, however, I think maybe putting it there will help communicate, so I'll do it.

PR #138, opened 2020-09-30
Short circuit on test assertion failures

philomates:
addresses https://github.com/nubank/state-flow/issues/21

I'm motivated to have this implemented because during service development, when I change something in an incorrect way, I'm often hit by several large blocks of stacktraces. Usually the first stack trace is the one I need to focus on, and the others, which arise from subsequent subflow executions, can be ignored until the first is addressed

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -19,9 +19,9 @@
             [changelog-check "0.1.0"]]
 
   :dependencies [[org.clojure/clojure "1.10.1"]
-                 [com.taoensso/timbre "4.10.0"]
-                 [funcool/cats "2.3.5"]
-                 [nubank/matcher-combinators "3.0.1"]]
+                 [com.taoensso/timbre "5.0.1"]
+                 [funcool/cats "2.3.6"]
+                 [nubank/matcher-combinators "3.1.3"]]
 
   :exclusions   [log4j]
 
@@ -33,9 +33,9 @@
   :profiles {:uberjar {:aot :all}
              :dev {:source-paths ["dev"]
                    :dependencies [[ns-tracker "0.4.0"]
-                                  [org.clojure/tools.namespace "0.3.1"]
+                                  [org.clojure/tools.namespace "1.0.0"]
                                   [midje "1.9.9"]
-                                  [org.clojure/java.classpath "0.3.0"]
+                                  [org.clojure/java.classpath "1.0.0"]
                                   [rewrite-clj "0.6.1"]]}}
 
   :aliases {"coverage" ["cloverage" "-s" "coverage"]
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject nubank/state-flow "5.5.0"
+(defproject nubank/state-flow "5.6.0"
   :description "Integration testing with composable flows"
   :url "https://github.com/nubank/state-flow"
   :license {:name "MIT"}
--- a/src/state_flow/assertions.clj
+++ b/src/state_flow/assertions.clj
@@ -1,0 +1,7 @@
+(ns state-flow.assertions)
+
+(defn failure?
+  "Is the result of a flow an assertion failure?"
+  [v]
+  ;; currently only matcher-combinator assertions are supported
+  (= :mismatch (:match/result v)))
--- a/src/state_flow/assertions.clj
+++ b/src/state_flow/assertions.clj
@@ -1,7 +1,0 @@
-(ns state-flow.assertions)
-
-(defn failure?
-  "Is the result of a flow an assertion failure?"
-  [v]
-  ;; currently only matcher-combinator assertions are supported
-  (= :mismatch (:match/result v)))
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -70,6 +70,12 @@
   []
   (state/gets state->current-description))
 
+(def fail-fast?
+  "Should the flow stop after the first failing assertion?
+
+  For internal use. Subject to change."
+  (state/gets (comp :fail-fast? meta)))
+
 (defn- clarify-illegal-arg [pair]
   (if-let [illegal-arg (some->> pair first :failure .getMessage (re-find #"cats.protocols\/Extract.*for (.*)$") last)]
     [(#'cats.monad.exception/->Failure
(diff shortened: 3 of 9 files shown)


src/state_flow/assertions.clj
```diff
@@ -0,0 +1,7 @@
+(ns state-flow.assertions)
+
+(defn failure?
+  "Is the result of a flow an assertion failure?"
+  [v]
+  ;; currently only matcher-combinator assertions are supported
+  (= :mismatch (:match/result v)))
```
philomates, line 7: this is a bit "hard-coded" but I guess is fine enough

src/state_flow/core.clj
```diff
@@ -224,18 +225,21 @@
...
            on-error               (comp throw-error!
                                         log-error
                                         (filter-stack-trace default-stack-trace-exclusions))}
     :as   opts}
    flow]
   (let [initial-state (init)
-        pair          (clarify-illegal-arg (runner flow
-                                                   (vary-meta initial-state assoc :runner runner)))]
+        pair          (clarify-illegal-arg
+                        (binding [state/*fail-fast?* fail-fast?]
```
philomates, line 240: using dynamic rebinding is the easiest way I could think to get this functioning, but if others can think of a way to implement it within the standard bounds of explicit control flow, I'm keen to here your suggestions

src/state_flow/core.clj
```diff
@@ -215,6 +215,7 @@
 
   Supported keys in the first argument are:
 
+    `:fail-fast?` optional, default `false`, when set to `true`, the flow stops running after the first failing assertion
```
philomates, line 218: one could also potentially leave this as a rebind that someone can call in the runner they define themselves and have them pass that runner in using the `:runner` command. It seemed a little to abstract even though it leaves the surface area of the api minimal

test/state_flow/core_test.clj
```diff
@@ -143,8 +143,21 @@
...
+                            {:runner (constantly (throw (ex-info "Oops" {})))}
+                            (state/get)))))
+
+  (testing "flow with fail-fast stops at first failing assertion"
+    (let [result (state-flow/run* {:init       (constantly {:value 0})
+                                   :fail-fast? true
+                                   :on-error   state-flow/ignore-error}
+                                  (flow "stop before boom"
+                                        (flow "mimic test failure manually"
+                                              (state/return {:match/result :mismatch}))
```
philomates, line 155: janky, but I didn't want to have to import `match?`. Maybe I should move this test to another ns and import `match?` to not have to mimic it manually with `:match/result :mismatch`?

project.clj
```diff
@@ -33,9 +33,9 @@
   :profiles {:uberjar {:aot :all}
              :dev {:source-paths ["dev"]
                    :dependencies [[ns-tracker "0.4.0"]
-                                  [org.clojure/tools.namespace "0.3.1"]
+                                  [org.clojure/tools.namespace "1.0.0"]
```
dchelimsky, line 36: Stability, at last!

src/state_flow/core.clj
dchelimsky: I think this the right place for it from an API perspective. It also leaves us the option to refactor the implementation not to use the dynamic var ;) (/me channeling @sovelten)

dchelimsky (approved):
LGTM. I do wonder if we could implement it passing the arg through instead of using a dynamic var, but I'm comfortable leaving it as/is and dealing with that in a subsequent refactoring, if at all.

Great addition IMO.

sovelten:
> addresses #21
> 
> I'm motivated to have this implemented because during service development, when I change something in an incorrect way, I'm often hit by several large blocks of stacktraces. Usually the first stack trace is the one I need to focus on, and the others, which arise from subsequent subflow executions, can be ignored until the first is addressed

I worked around that by using the command `cider-repl-previous-prompt` on emacs that will jump to the point the error reports started :D

src/state_flow/state.clj
```diff
@@ -33,9 +36,14 @@
       (error-catching-state
        (fn [s]
          (let [[v s'] ((p/-extract fv) s)]
-           (if (e/failure? v)
-             [v s']
-             [(result-or-err f v) s'])))))
+           (cond (e/failure? v)
```
sovelten, line 39: I don't think this is the right abstraction level to implement this change.
We should be storing assertion results in the meta of the state (if not, we should be doing that with the new `match?`)
We can choose to throw/stop if the assertion being stored is a failed assertion or something like that.
This will result in an incomplete test report. Maybe we should be aiming to stop outputting failures instead of stopping the test.

sovelten: If you store this option of the state meta, `match?` can access fetch the option from the meta and throw after a test fails. Throwing will shortcut the execution. I think this is simple and effective. WDYT?

philomates: I didn't want achieve shortcircuiting via raising an exception because then we would be mixing that special behavior with the normal exception handling flow. Doing so would mean that when shortcircuiting is enabled a test failure would also appear as an exception, right? Does this seem like a valid thing to be concerned about mixing?

The idea of continuing to run the tests but only reporting the first failure feels a little strange to me. Like what if the subsequent steps are kinda slow? Or what if you want to leave the system at the exact state that it was at after the first assertion failure so that you can inspect it?

That said, storing the flag in the meta of the state sounds cleaner than using `binding`. I can look to see if that comes together well

sovelten: Technically, the state monad implementation will short circuit on `failure?` as defined here: https://github.com/nubank/cats/blob/master/src/cats/monad/exception.cljc#L151

So having `match?` return a failure when fail-fast is turned on should also short circuit the run.
However, the error-handler probably needs to distinguish between this and other exceptions if we don't want to just throw it.

philomates: Ah, right; I had considered that in the beginning but decided it would be better to avoid encoding a "special exception" that we treat differently. In the end my implementation does the exact same things with a "special return value" that stops further execution. So I guess it comes down to if we want to hijack the exceptions control flow or the normal control flow. I suppose since exceptions always stop execution, using that flow is a little more in-line, and then the only exceptional case we encode is whether we throw the exception or swallow it (for test assertion failures) in the end.

So, all that said, I should switch the implementation to use a special exception and toggle the short-circuit behavior via a meta flag in the state?

sovelten: That looks good to me :)
Just to clarify, I don't think you need to throw an exception, just return a failure data structure should be equivalent. I don't think I ever tested this though :thinking: 

Let me know if you want help with that.

philomates:
@sovelten @dchelimsky I've redone the implementation based on https://github.com/nubank/state-flow/pull/138#discussion_r497588587 to use state metadata and exception short-circuiting to implement the feature. Could you take a look again?

sovelten (approved):
Looks good to me :tada:

sovelten:
Might be good to add this info on the README somewhere

PR #142, opened 2020-10-29
add flow/for macro

dchelimsky:
## Some existing solutions:

* defn a fn that returns a flow and invoke it from a defflow

```clojure
(defn a-flow [arg] (flow (str "do something with " arg) ,,,)

(defflow do-something-with-x (a-flow 'x))
(defflow do-something-with-y (a-flow 'y))
;; or
(defflow do-something-with-x-and-y
  (a-flow 'x)
  (a-flow 'y))
```

* use cats.core/sequence directly

```clojure
(defflow do-something-with-x-and-y
  (cats.core/sequence
    (for [z [:x :y]]
      (flow (str "do stuff with " z)
       ,,,
       ))))
```

## Proposal

This PR adds a `state-flow.api/for` macro that works just like `clojure.core/for`, but returns a flow which wraps a sequence of flows, e.g.

```clojure
(defflow do-something-with-x-and-y
  (flow/for [z [:x :y]]
    (flow (str "do stuff with " z)
     ,,,
     ))))
```

TODO:
* [x] more doc

### DIFF
--- a/src/state_flow/api.clj
+++ b/src/state_flow/api.clj
@@ -1,5 +1,7 @@
 (ns state-flow.api
-  (:require [state-flow.assertions.matcher-combinators]
+  (:refer-clojure :exclude [for])
+  (:require [cats.core :as m]
+            [state-flow.assertions.matcher-combinators]
             [state-flow.cljtest]
             [state-flow.core]
             [state-flow.state]
@@ -38,3 +40,22 @@
 
 (import-fn state-flow.state/gets   get-state)
 (import-fn state-flow.state/modify swap-state)
+
+(defmacro for
+  "Like clojure.core/for, but returns a flow which wraps a sequence of flows e.g.
+
+     (flow \"even? returns true for even numbers\"
+       (flow/for [x (filter even? (range 10))]
+         (match? even? x)))
+
+     ;; same as
+
+   (flow \"even? returns true for even numbers\"
+     (match? even? 0)
+     (match? even? 2)
+     (match? even? 4)
+     (match? even? 6)
+     (match? even? 8)) "
+  [seq-exprs flow]
+  `(m/sequence
+    (clojure.core/for ~seq-exprs ~flow)))
--- a/README.md
+++ b/README.md
@@ -104,6 +104,28 @@ a system using [Stuart Sierra's Component](https://github.com/stuartsierra/compo
 (state-flow.api/run* {:init (constantly {:service-system (atom nil))} flow)
 ```
 
+### Composing Flows
+
+Flows follow the Composite Pattern: a single flow has the same
+interface as a collection of flows.
+
+You can compose flows by nesting them in other flows:
+
+``` clojure
+(flow "do many things"
+  (flow "do one thing" ,,,)
+  (flow "do another thing" ,,,))
+```
+
+Use `state-flow.api/for` when you have a flow that you'd like to apply
+to different inputs with the same outcome, e.g.
+
+``` clojure
+(flow "even? returns true for even numbers"
+  (flow/for [x (filter even? (range 10))]
+    (match? even? x)))
+```
+
 #### Failing Fast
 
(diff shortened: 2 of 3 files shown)


sovelten:
I don't think I would be a user of this macro, but I understand if people want to use it.

I do think this is general enough to be part of https://github.com/nubank/cats/blob/master/src/cats/core.cljc though. It works for any monad, so it is kind of syntactic sugar for `m/sequence`

dchelimsky:
> I don't think I would be a user of this macro, but I understand if people want to use it.
> 
> I do think this is general enough to be part of https://github.com/nubank/cats/blob/master/src/cats/core.cljc though. It works for any monad, so it is kind of syntactic sugar for `m/sequence`

I'm happy to submit a separate PR for cats, but I also like the idea of having this in state-flow so it's obvious to state-flow users.

dchelimsky:
> it is kind of syntactic sugar for `m/sequence`

I agree, but I think that `(m/sequence (for [,,,] many-lines-of-flows))` is easier to scan than `(m/sequence (map (fn [,,,] many-lines-of-flows) seq-of-args)`, so I want to encourage `for` for consistency.

dchelimsky:
> > I don't think I would be a user of this macro, but I understand if people want to use it.
> > I do think this is general enough to be part of https://github.com/nubank/cats/blob/master/src/cats/core.cljc though. It works for any monad, so it is kind of syntactic sugar for `m/sequence`
> 
> I'm happy to submit a separate PR for cats, but I also like the idea of having this in state-flow so it's obvious to state-flow users.

https://github.com/funcool/cats/pull/241

src/state_flow/api.clj
```diff
@@ -38,3 +40,22 @@
 
 (import-fn state-flow.state/gets   get-state)
 (import-fn state-flow.state/modify swap-state)
+
+(defmacro for
```
sovelten, line 44: should this be implemented here or in another namespace, such as in `state-flow.state`?

sovelten (approved):

src/state_flow/api.clj
dchelimsky: It's not really a `state` fn, but it's not really `core` for state-flow either. We could add a new namespace like `state-flow.monad`, but that might be confusing/distracting. Other ideas?

What about `state-flow.compose`? That might encourage thinking about some other ideas for composing flows.

FelipeCortez:
> > I don't think I would be a user of this macro, but I understand if people want to use it.
> > I do think this is general enough to be part of https://github.com/nubank/cats/blob/master/src/cats/core.cljc though. It works for any monad, so it is kind of syntactic sugar for `m/sequence`
> 
> I'm happy to submit a separate PR for cats, but I also like the idea of having this in state-flow so it's obvious to state-flow users.

not very long ago I struggled when migrating a test because there was nothing out-of-the-box like this in state-flow, and using cats like the documentation suggested felt a bit wrong, so +1. I think this is also in line with the idea that you don't have to know about the state monad to use state-flow

src/state_flow/api.clj
FelipeCortez: I vote for `state-flow.compose`

sovelten: the `state` is actually a `state monad` so there should be the place IMO, it is in the same spirit of `fmap`. Not that I have a strong opinion about it.

FelipeCortez (approved):

src/state_flow/api.clj
sovelten: In case you want to use this helper in core.clj, core.clj depends on the state.clj, but it should not depend on state-flow.api, the hierarchy looks like this:

`state`
|
`core`
|
`api`

things that are more fundamental than core go in the state

dchelimsky: The struggle I'm having is that e.g. `swap-state` is clearly about `state`, whereas `for` is about composing flows. I've already merged this but we can always move it somewhere else later and reference it from `api` as we do with all the other functions.

dchelimsky: > whereas `for` is about composing flows.

And composing flows seems more fundamental than even `state` to me :)

Issue #136, opened 2020-08-24
Allow customization of the metadata handler fn in the runner


sovelten:
See https://github.com/nubank/visual-flow/issues/27 for motivation
(ie. allow external tools to have access to flow description metadata at runtime)

gabrielgiussi:
@sovelten Remember that visual-flow is internal

PR #144, opened 2020-11-03
Add before flow hook

sovelten:
This change enables the user to add a hook that is executed before flow execution and after description is updated.

One example of use case is allowing access from third-party tools of the description stack at run-time

Solves https://github.com/nubank/state-flow/issues/136

### DIFF
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -86,7 +86,7 @@
 (defn apply-before-flow-hook
   []
   (m/do-let
-   [hook (state/gets #(-> % meta :before-flow-hook))]
+   [hook (state/gets (comp :before-flow-hook meta))]
    (state/modify (or hook identity))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/src/state_flow/core.clj
+++ b/src/state_flow/core.clj
@@ -83,6 +83,12 @@
      (second pair)]
     pair))
 
+(defn apply-before-flow-hook
+  []
+  (m/do-let
+   [hook (state/gets #(-> % meta :before-flow-hook))]
+   (state/modify (or hook identity))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Public API
 
@@ -103,6 +109,7 @@
         flows'    (or flows `[(m/return nil)])]
     `(m/do-let
       (push-meta ~description ~flow-meta)
+      (apply-before-flow-hook)
       [ret# (m/do-let ~@flows')]
       (pop-meta)
       (m/return ret#))))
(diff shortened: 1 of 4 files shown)


src/state_flow/core.clj
```diff
@@ -83,6 +83,12 @@
      (second pair)]
     pair))
 
+(defn apply-before-flow-hook
+  []
+  (m/do-let
+   [hook (state/gets #(-> % meta :before-flow-hook))]
```
dchelimsky, line 89: 💅 
```suggestion
   [hook (state/gets (comp :before-flow-hook meta))]
```

test/state_flow/core_test.clj
```diff
@@ -311,3 +311,13 @@
       (is (empty? (->> (rest frames)
                        (map #(.getClassName %))
                        (filter #(re-find #"^clojure.lang" %))))))))
+
+(deftest before-flow-hook
+  (testing "default: uses default-stack-trace-exceptions (on all but first frame)"
+    (let [result (->> (state-flow/run*
+                       {:before-flow-hook (fn [s] (assoc s :my-thing (str (first (clojure.string/split (#'state-flow/state->current-description s)
```
dchelimsky, line 318: I'm not sure about the name `:before-flow-hook`. To me, the "hook" is the mechanism to which you provide a function of state, so maybe `:before-flow-fn`? Also, I wonder whether this hook should accept a flow, so this would be

```clojure
{:before-flow (flow/swap-state assoc :my-thing k v)) ,,,
```

That would give users access to more tools. WDYT?

sovelten: We thought about it (cc @Andre0991), but this would probably cause the flow macro to enter an infinite loop (because the `:before-flow` of the before-flow is also a flow).

For me a hook is something that you attach to something else, in this case we are hooking a function to be executed whenever the flow runs. But I don't mind changing to `:before-flow-fn`, I'm not aware of differing interpretations of hook.

sovelten: actually it would be a runtime loop because we are fetching the flow at execution time

dchelimsky: In that case, how about naming it something like `:swap-state-before` so it's clearer how the function is used?

sovelten: `:before-flow-state-modifier-fn`? We settled on `state/modify`, so I think modifier is a better fit.

It doesn't necessarily need to modify it, though, it can be used to access the current state ^^^^^^^dynamically^^^^^^^^^^

dchelimsky: > We settled on state/modify, so I think modifier is a better fit.

We had settled on `state/modify`, but later settled on [`flow/swap-state` ](https://github.com/nubank/state-flow/blob/fbfd0469e3a30bedc049ee9d26355ab4cb713595/src/state_flow/api.clj#L42) and `flow/get-state`.

The implementation is calling `state/modify` a.k.a. `flow/swap-state`, and I think it's useful to declare that.

> It doesn't necessarily need to modify it, though

That's the reason I want this to be a flow. Then you could bind any of the primitives. I wonder if there's a different implementation we could use to solve that.

sovelten: `:before-flow-swap-state-fn`?

I don't think there is a way around it, so it's safer to ask the user for a regular function and wrap it around `swap-state`.
Accepting suggestions though

sovelten: you could wrap the passed flow in some other flow with a unique identifier description that might be used to break the runtime loop, as we would be able to recognize the flow being executed as one that should not have a flow applied to it. It doesn't look very pretty though.

sovelten: We could wrap the flow we receive with a flow that sets the `:before-flow-swap-state-fn` with a no-op flow, than reset to the original after it. That will do it in a elegant way. Makes sense?

dchelimsky: I'm working on something ... coming soon ...

dchelimsky: > We could wrap the flow we receive with a flow that sets the `:before-flow-swap-state-fn` with a no-op flow, than reset to the original after it. That will do it in a elegant way. Makes sense?

Makes sense to me!

sovelten: Another problem is that this flow should not push and pop descriptions to meta either or we will have bogus descriptions

dchelimsky: I ran into that exact problem in my experiments!

I'm happy to let this go as/is.

dchelimsky (approved):

test/state_flow/core_test.clj
sovelten: Any naming preferences?

dchelimsky: I'd say either what you have already, `:before-flow-hook`, or just `:before-flow`.

Andre0991 (approved):

PR #145, opened 2020-11-04
Add filename to flow descriptions

Andre0991:
- Add filename to flow descriptions. Use the following format: `<description> (<file>:<line>)`
- Put a map with flow metadata (line, file, etc) in state instead of directly putting the formatted string.

### DIFF
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -59,7 +59,7 @@
                             :caller-meta  caller-meta
                             :times-to-try 1
                             :sleep-time   probe/default-sleep-time}
-                                params)]
+                           params)]
 
     (core/flow*
      {:description (:description params*)
--- a/src/state_flow/assertions/matcher_combinators.clj
+++ b/src/state_flow/assertions/matcher_combinators.clj
@@ -52,11 +52,13 @@
    ;; deprecated cljtest/match? fn.  Undecided
    ;; whether we want to make it part of the API.
    ;; caller-meta is definitely not part of the API.
-  (let [caller-meta      (meta &form)
-        params*          (merge {:description  "match?"
-                                 :caller-meta  caller-meta
-                                 :times-to-try 1
-                                 :sleep-time   probe/default-sleep-time}
+  (let [f           *file*
+        a-ns        (str *ns*)
+        caller-meta (assoc (meta &form) :file f :ns a-ns)
+        params*     (merge {:description  "match?"
+                            :caller-meta  caller-meta
+                            :times-to-try 1
+                            :sleep-time   probe/default-sleep-time}
                                 params)]
 
     (core/flow*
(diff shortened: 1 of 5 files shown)


src/state_flow/core.clj
```diff
@@ -31,13 +31,16 @@
...
-  [description {:keys [line]}]
+  [description {:keys [line ns file]}]
   (modify-meta
    (fn [m] (-> m
                (update :top-level-description #(or % description))
-               (update :description-stack (fnil conj []) (str description
-                                                              (when line
-                                                                (format " (line %s)" line))))))))
+               (update :description-stack (fnil conj [])
+                       ;; TODO: Add filename as well
```
sovelten, line 39: remove todo

src/state_flow/assertions/matcher_combinators.clj
```diff
@@ -52,12 +52,14 @@
...
    ;; caller-meta is definitely not part of the API.
-  (let [caller-meta      (meta &form)
-        params*          (merge {:description  "match?"
-                                 :caller-meta  caller-meta
-                                 :times-to-try 1
-                                 :sleep-time   probe/default-sleep-time}
-                                params)]
+  (let [f           *file*
+        a-ns        (str *ns*)
+        caller-meta (assoc (meta &form) :file f :ns a-ns)
```
dchelimsky, line 57: 💅 I'd prefer inlining e.g. `(assoc (meta &form) :file *file* :ns *ns*)`, and maybe even inlining the caller-meta in params* below.

src/state_flow/core.clj
```diff
@@ -31,13 +31,15 @@
...
-  [description {:keys [line]}]
+  [description {:keys [line ns file]}]
   (modify-meta
    (fn [m] (-> m
                (update :top-level-description #(or % description))
-               (update :description-stack (fnil conj []) (str description
-                                                              (when line
-                                                                (format " (line %s)" line))))))))
+               (update :description-stack (fnil conj [])
+                       (merge {:description description
```
dchelimsky, line 39: I know how big a fan of `cond->` you are, so I'll just throw this out as a suggestion with a weak preference:

```clojure
(cond-> {:description description :ns ns)
  line (assoc :line line)
  file (assoc :file file))
```

src/state_flow/core.clj
```diff
@@ -118,9 +129,13 @@
...
   {:style/indent :defn}
   [description & flows]
-  (apply flow* {:description description
-                :caller-meta (meta &form)}
-         flows))
+  (let [f    *file*
+        a-ns (str *ns*)]
+    (apply flow* {:description description
+                  :caller-meta (assoc (meta &form)
+                                      :file f
```
dchelimsky, line 136: 💅 inline `f` and `a-ns`

dchelimsky (approved):

src/state_flow/core.clj
sovelten:
```clojure
(assoc-some {:description description :ns ns}
  :line line
  :file file)
```
would be the ideal solution in my opinion, but not part of the core. I have my reasons for avoiding arrow macro shenanigans, but will let @Andre0991 decide :P

sovelten (approved):

PR #147, opened 2020-12-22
Add `state-flow.state/when`

Andre0991:


src/state_flow/state.clj
```diff
@@ -110,6 +110,13 @@
   [my-fn]
   (error-catching-state (fn [s] [(my-fn) s])))
 
+(defn when
```
dchelimsky, line 113: who? what? where? why? how?

Seriously - what problem is this solving?

Andre0991: It's equivalent to cats.core's when (https://funcool.github.io/cats/latest/api/cats.core.html#var-when) that works with flows.

It makes it a bit more convenient to make a flow run only if a condition is met.

The cats.core version won't work with flows because it doesn't implement some protocol or something like that - @sovelten knows the gory details.

sovelten: We need to implement a protocol for applicative on the State record for `cats.core/when` to work.
Meanwhile we can have this available on the api

Instead of writing `(if a b (state/return nil))` we write `(when a b)`

sovelten: why cats.core/when is a macro is probably not to evaluate b if a is falsey.

sovelten: (which doesn't make a lot of sense in this case because the state record is a frozen evaluation anyway, so you just refrain from evaluating the evaluation that will create the frozen evaluation)

philomates (approved):
:+1: I've found myself wanting this a time or two

src/state_flow/state.clj
Andre0991: @dchelimsky, WDYT?

sovelten (approved):

src/state_flow/state.clj
dchelimsky: I just don't understand the purpose of this in a test. I'm imagining "generate some states and run the test if the generated state is testable". In that case, fix the generator. I'm sure I'm missing something though. What is it?

Andre0991: Not sure if I get what you are saying 🤔

`(when a b)` is just a shortcut for `(if a b (state/return nil))`, which would be handy sometimes.

dchelimsky: Don't let me get in the way of adding this. I'm not against its existence, just its abuse. If I see such abuse, I'll take it up with the abuser :)

sovelten: This is an example of code on state-flow itself 
https://github.com/nubank/state-flow/pull/70/commits/2fff6c90f5797b1dfd0c62d6f9b43489c0896bd6#diff-5be8b2b2ad4afc021a37f1f169fa0f6fcd78e1c5ff75511ed72108f152d14e4bR88
It didn't end up being pushed because we managed to find a better solution, but that's the idea.

dchelimsky: For clarity - the scenario I imagine somebody using is e.g.

```clojure
(defflow " .... "
  [thing (generate-a-thing)]
  (flow/when (thing-is-in-a-valid-state thing)
    (flow "run these tests for valid states" ,,,))
  (flow/when (not (thing-is-in-a-valid-state thing))
    (flow "run these other tests for invalid states" ,,)))
```

I'll bet you a beer this happens, to be paid the next time I'm in Brazil after it does :)

philomates: I need it here: https://github.com/nubank/common-multi-service-tests/pull/64/files#diff-490d11171c7b806f5ae93528cda19c7809083d524e6241ee90b8862f6f0f7609L49

philomates: https://github.com/nubank/state-flow/pull/147#discussion_r574679551

this can be achieved with an `if` statement also, right?

Andre0991: Oh, got it.

Since you are comfortable with adding the function, I'm releasing this.

philomates:
heads up that I just released this as `5.11.0`

PR #149, opened 2021-03-04
Add clj-kondo hook to defflow

ericdallo:
The clj-kondo hook should tell to clj-kondo how to lint the custom macro `defflow`

Ref 1: https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting-and-importing-configuration
Ref 2: https://github.com/clj-kondo/clj-kondo/blob/master/doc/hooks.md#hooks

clj-kondo will find from the classpath the hook

With that we can just add ` :config-paths ["nubank/state-flow"]` to our clj-kondo home config and everything should work :)


thiagokokada (approved):
Do we need to update `state-flow` in all services to get it to work :thinking: ?

ericdallo:
Yes @thiagokokada as we get from the classpath that config

PR #150, opened 2021-03-09
Question: should exceptions raised in `match?` respect :times-to-try or short-circuit?

philomates:
some code we wrote expects retries (specified by the `:times-to-try`) to happen for sub-flows that raise exceptions. Turns out the error short-circuiting state monad implementation we use (I think) causes exceptions to not lead retries. This PR has a failing test case that exemplifies this.

Should we address this or is it beneficial to leave the behavior as is?
If we change this I imagine it would be a breaking change.


philomates:
hmm, I think this might be the expected/desired behavior. Perhaps I'll update my test to instead demonstrate this behavior to document that this is expected

PR #151, opened 2021-03-18
Add clj-kondo defined-by to defflow hook

ericdallo:
Add `:defined-by` in `defflow` clj-kondo hook used by latest clj-kondo. This should fix clojure-lsp warnings about `unused-public-var` linter.


leoiacovini (approved):
🎉

PR #152, opened 2021-04-08
Add description-stack to the API

Andre0991:
This information is useful for extracting data from tests. This enables
us (for example) to generate visualizations with descriptions. See https://github.com/nubank/visual-flow/pull/36

### DIFF
--- a/src/state_flow/api.clj
+++ b/src/state_flow/api.clj
@@ -25,13 +25,14 @@
          when
          description-stack)
 
+;; The vars below represent the user API, that is,
+;; the set of public functions that one can use for writing flows.
 (import-vars
  state-flow.core/flow
  state-flow.core/run
  state-flow.core/run*
  state-flow.core/log-and-throw-error!
  state-flow.core/ignore-error
- state-flow.core/description-stack
 
  state-flow.state/invoke
  state-flow.state/return
--- a/src/state_flow/api.clj
+++ b/src/state_flow/api.clj
@@ -23,8 +23,7 @@
          match?
          get-state
          swap-state
-         when
-         description-stack)
+         when)
 
 (import-vars
  state-flow.core/flow
--- a/src/state_flow/api.clj
+++ b/src/state_flow/api.clj
@@ -1,4 +1,5 @@
 (ns state-flow.api
+  "This namespace provides the user API, that is, the set of public functions that one can use for writing flows."
   (:refer-clojure :exclude [for when])
   (:require [cats.core :as m]
             [state-flow.assertions.matcher-combinators]
@@ -25,8 +26,6 @@
          when
          description-stack)
 
-;; The vars below represent the user API, that is,
-;; the set of public functions that one can use for writing flows.
 (import-vars
  state-flow.core/flow
  state-flow.core/run
--- a/src/state_flow/api.clj
+++ b/src/state_flow/api.clj
@@ -22,7 +22,8 @@
          match?
          get-state
          swap-state
-         when)
+         when
+         description-stack)
 
 (import-vars
  state-flow.core/flow
@@ -30,6 +31,7 @@
  state-flow.core/run*
  state-flow.core/log-and-throw-error!
  state-flow.core/ignore-error
+ state-flow.core/description-stack
 
  state-flow.state/invoke
  state-flow.state/return
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [5.12.0]
-* Make `state-flow.core/description-stack` public.
+* Label `state-flow.core/description-stack` as public (was only for internal usage).
 
 ## [5.11.3]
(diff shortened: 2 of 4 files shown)


src/state_flow/api.clj
```diff
@@ -22,14 +22,16 @@
...
+         when
+         description-stack)
 
 (import-vars
  state-flow.core/flow
  state-flow.core/run
  state-flow.core/run*
  state-flow.core/log-and-throw-error!
  state-flow.core/ignore-error
+ state-flow.core/description-stack
```
sovelten, line 34: I would not put it here as part of the API

I understand this is the user API
`state-flow.core/description-stack` is part of developer API

Andre0991: I thought the set of all those vars represented the public interface of `state-flow`, nothing more.

Do we have more examples of other functions that are part of the 'developer API', but not core?

Andre0991: My concern is how to make it visible that `description-stack` is part of the API and shound't have breaking changes.

sovelten: Everything that is public and is not commented as "internal use only" is part of the 'developer API'
This is not very formal though, but it is assumed that changes on public functions that are not marked as internal are breaking changes.

dchelimsky: Knee-jerk thought (I haven't thought this through): how about a separate namespace `state-flow.developer-api`?

sovelten: Are we mature enough for a formal developer API? We can certainly work on that but not sure this is the time.

fernando-nubank: Here we go again :) What defines the _public_ API of a library?
Our (probably unwritten) agreement for state-flow related libs was that everything public is an API, except when noted on the docstring.

The namespace `state-flow.api` was created to simplify the _use_ of id when writing tests.

dchelimsky: Agreed it's premature to introduce a new namespace. I'm just thinking of how to address @Andre0991 's concern. I think @sovelten may have done that already, so never mind me.

Andre0991: IMO, marking stuff as part of the API explicitly is easier to reason about compared to "everything public is part of the API unless its docstring say otherwise"

So I like the idea of having a `developer-api` ns.

But I suppose you have already discussed this extensively, so I will respect the current convention.

Andre0991: > Our (probably unwritten) agreement for state-flow related libs was that everything public is an API, except when noted on the docstring.

Should we add a comment to state-flow.api explaining that?

sovelten: You can add a comment saying state-flow.api is the recommended user API

Andre0991: Done. Please re-review.

dchelimsky: It's all over the readme (and with https://github.com/nubank/state-flow/pull/153, even more so).

src/state_flow/api.clj
```diff
@@ -22,8 +22,11 @@
          match?
          get-state
          swap-state
-         when)
+         when
+         description-stack)
 
+;; The vars below represent the user API, that is,
```
dchelimsky, line 28: I think this would be better in a docstring in the namespace declaration.

Andre0991: Agreed, done

sovelten (approved):

dchelimsky:
Wait - I'm sorry, but I'm confused. I don't think description-stack belongs on the API. More coming ...

dchelimsky:
> Wait - I'm sorry, but I'm confused. I don't think description-stack belongs on the API. More coming ...

I agree that we want a way to assure developers of tooling that they can rely on description-stack, and I also agree that it's too early to formalize a developer API, but I don't see it as part of normal use in writing flows.

I'd rather have a way to formalize that changes to any function not explicitly labeled "internal use" would be considered a breaking change.

Andre0991:
I agree, actually. All we want is to be able to rely on it, it does not have the be part of the API.

In any case, it makes sense for this function to be public, right?

We can discuss how to formalize the 'dev api' better, but for now, do you think it's OK to merge this PR as it is?

dchelimsky:
> @dchelimsky, we can discuss how to formalize the 'dev api' better, but for now, do you think it's OK to merge this PR as it is?

I'm not comfortable including it in the `state-flow.api` namespace. I'd rather expand the docstring for description-stack indicating that we consider it a public function for developer use. But I can't stop you from merging :)

sovelten:
I think @Andre0991 just forgot to remove it from the api, it shouldn't be included as we discussed

Andre0991:
Oh, I forgot to remove it from the `declare` as well 🤦  Sorry

dchelimsky:
> I think @Andre0991 just forgot to remove it from the api, it shouldn't be included as we discussed

> Oh, I forgot to remove it from the declare as well 🤦 Sorry

That's how I got confused :)

PR #154, opened 2021-04-22
recursively filter stack-traces + shorten other stack traces

philomates:
If people like the changes I'll write tests for them

### before

```
2021-04-22T18:04:45.468Z mboi-tui INFO [state-flow.core:8] - Flow "run-main-locally" failed with exception
    nrepl.middleware.interruptible-eval/evaluate/fn     interruptible_eval.clj:   87
                                                ...
                        clojure.core/with-bindings*                   core.clj: 1977 (repeats 2 times)
                                 clojure.core/apply                   core.clj:  667
                                                ...
 nrepl.middleware.interruptible-eval/evaluate/fn/fn     interruptible_eval.clj:   87
                                  clojure.core/eval                   core.clj: 3202
                                                ...
                              user$eval58135.invoke             NO_SOURCE_FILE:    1
                        user$eval58135.invokeStatic             NO_SOURCE_FILE:    1
                                                ...
                             clojure.test/run-tests                   test.clj:  768 (repeats 2 times)
                                 clojure.core/apply                   core.clj:  669
                                                ...
                                clojure.core/map/fn                   core.clj: 2759
                               clojure.test/test-ns                   test.clj:  758
                         clojure.test/test-all-vars                   test.clj:  737
                             clojure.test/test-vars                   test.clj:  731
                       clojure.test/default-fixture                   test.clj:  687
                          clojure.test/test-vars/fn                   test.clj:  735
                       clojure.test/default-fixture                   test.clj:  687
                       clojure.test/test-vars/fn/fn                   test.clj:  735
                              clojure.test/test-var                   test.clj:  717
                           clojure.test/test-var/fn                   test.clj:  717
               integration.apply-migrations-test/fn  apply_migrations_test.clj:  171
                               state-flow.core/run*                   core.clj:  299
                                state-flow.core/run                   core.clj:  265
                               cats.monad.state/run                 state.cljc:  147
           state-flow.state/error-catching-state/fn                  state.clj:   20
                                                ...
                       cats.monad.exception/wrap/fn             exception.cljc:  250
                   cats.monad.exception/exec-try-on             exception.cljc:  197
          cats.monad.exception/wrap/fn/func--auto--             exception.cljc:  250
                                 clojure.core/apply                   core.clj:  667
                                                ...
                          state-flow.state/reify/fn                  state.clj:   50
           state-flow.state/error-catching-state/fn                  state.clj:   20
                                                ...
                       cats.monad.exception/wrap/fn             exception.cljc:  250
[...]

PR #155, opened 2021-05-13
unbump timbre, bump midje + clojure

philomates:
https://github.com/nubank/state-flow/pull/154/files#r631432230


dchelimsky (approved):
Thanks! That resolves it.

PR #163, opened 2022-01-31
Fix state-flow pretty-printing for :actual in test result

michaelweisser:
image

Due to a change in cider-nrepl ( https://github.com/clojure-emacs/cider-nrepl/commit/dff71aade8a8e73b0350f9ecb0f4a36cfb568d85 ), emacs cider wasn't printing the "actual" part of a matcher combinator result correctly. This can be seen on the right side of the screen shot. Cider includes `:summary ...` which shouldn't appear. The output as it should be can be seen on the left.

With this patch, we are using matcher-combinator's meta tag for mismatches. This way, cider-nrepl will use the dispatching print function and the "actual" section of the test result will be pretty-printed as usual.

Fixes https://nubank.atlassian.net/browse/DEV-471

### DIFF
--- a/src/state_flow/cljtest.clj
+++ b/src/state_flow/cljtest.clj
@@ -1,6 +1,8 @@
 (ns state-flow.cljtest
   (:require [clojure.test :as t]
-            [matcher-combinators.printer :as matcher-combinators.printer] ;; to register clojure.test assert-expr for `match?`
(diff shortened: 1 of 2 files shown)


ericdallo (approved):

lednubr:
Same error ocurring (before the patch) in calva/vsscode

image

PR #180, opened 2024-12-12
Defflow in tutorial

kdmoreira:
The tutorial inside samples contains a 'Tests' section, where `state-flow.api/match?` is presented.

This PR adds examples featuring `state-flow.api/defflow` as well, since it can report mismatched assertions within flows, whereas `match?` will only return the result of the last evaluation.


sovelten (approved):
can you add a link to the tutorial in the README?

PR #185, opened 2025-03-05
Use setMacro to flag macros in the api namespace

carloshernandez2:
**Problem**

It is not possible to AOT-compile namespaces that use the defflow macro from the API namespace and get a faster development workflow as described [here](https://clojure.org/guides/dev_startup_time).

**Description**

Taking these namespaces as a starting point:

```clojure
(ns death-by-compilation-clj.my-defflow
  (:require [clojure.test :refer :all]))

(defmacro defflow
  [name]
  `(def ~name 1))
```
```clojure
(ns death-by-compilation-clj.my-dear-defflow
  (:require [clojure.test :refer :all]
            [death-by-compilation-clj.my-defflow :as my-defflow]))

(def ^{:macro    true}
  defflow #'my-defflow/defflow)
```
```clojure
(ns death-by-compilation-clj.my-dear-test
  (:require [clojure.test :refer :all]
            [death-by-compilation-clj.my-dear-defflow :refer [defflow]]))

(defflow my-dear-test-test)
```

When running `compile` on `death-by-compilation-clj.my-dear-test` it successfully compiles and writes to disk. Then if I kill the REPL, modify the file and repeat the process, the following error is thrown:

```
Syntax error compiling at (death_by_compilation_clj/my_dear_test.clj:7:1).
Unable to resolve symbol: my-dear-test-test in this context
```

**Solution**

By calling the setMacro method on the Vars that correspond to macros in the API namespace, we can flag them as the compiler expects and also add the macro metadata.

**References**

- https://clojurians-log.clojureverse.org/clojure/2023-01-18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
+## [5.20.1]
+- Use setMacro to flag macros in the api namespace to avoid AOT compilation errors
+
 ## [5.20.0]
 - Remove `com.taoensso/timbre` dependency
-  - **BEHAVIOR CHANGE**: previously, failures logged by `state-flow.core/log-error` were sent to whatever 
+  - **BEHAVIOR CHANGE**: previously, failures logged by `state-flow.core/log-error` were sent to whatever
     appenders were configured in timbre. Now, failures are always logged to `clojure.core/*out*`.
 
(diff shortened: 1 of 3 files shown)


acamargo (approved):
Nice!

